PR #8, opened 2018-03-06
Add SDKs for WPF and Windows Forms

wjk:
This PR adds SDKs that can be used to enable build and designer support for SDK-style projects that target the .NET Framework and use WPF or Windows Forms for their user interface. README files containing instructions for use are included.

I submitted these SDKs as a PR to this repository because I think they should use the `Microsoft.NET.Sdk.*` naming convention, and that requires that they be published through a Microsoft-operated NuGet account. Additionally, I would be appreciative if the SDK packages could be pushed to the repository's MyGet feed as soon as possible after this PR is merged, so I can start using them in my projects. Thanks so much!


jeffkl:
At this time, we don't think we can accept this.  A lot of the IP appears to be from onovotny but also repackaged official targets.  We'd like to involve the internal Microsoft teams that own these project systems to get an official solution out.  I'm fine with keeping this third party under a different name.

Can you please open an issue (or one for each project type) so we can track the work as we get more people involved?

wjk:
@jeffkl That's fine. I just thought the naming convention would work by analogy to `Microsoft.NET.Sdk.Web` and so on. I am perfectly fine with renaming them and publishing them from my own repository; in fact, that would help simplify my deployments quite a bit. I will post links to the renamed packages soon. Thanks nevertheless!

jeffkl:
Thanks for your understand and willingness to contribute none the less.  I want to provide a great first party solution and hopefully its as simple as just having you contribute back what you have after I get some stakeholders involved.

PR #31, opened 2018-05-23
Version 2.0 of central package versions

jeffkl:
Use `<PackageReference Update="" />` instead of `<PackageVersion />` because having two lists is hard to get metadata right.

Update README


jeffkl:
1. Brainstorm a better say of handling global references, maybe just `DevelopmentDependency="true"` in NuGet proper which is a shortcut for `PrivateAssets="All` and `ExcludeAssets="Compile;Runtime"`
2. Figure out how to ship this officially
  a. Can CPS write to `Packages.props`?
  b. Are NuGet clients involved in any way?
  c. Who would ship the target?

PR #54, opened 2018-09-28
Add support for command `dotnet test` and `dotnet publish`

yhvicey:
This change enables this sdk to execute `VSTest` target and `Publish` in referenced project.

However, If a dotnet project is not a test project, it will also has a `VSTest` target, which will lead to a build error. Maybe need to add some switch to control it.

### DIFF
--- a/src/Traversal/Sdk/Sdk.targets
+++ b/src/Traversal/Sdk/Sdk.targets
@@ -63,6 +63,10 @@
     <PackDependsOn>
       ResolveReferences;
     </PackDependsOn>
+
+    <PublishDependsOn>
+      Build;
+    </PublishDependsOn>
   </PropertyGroup>
 
   <ItemGroup Condition=" '$(TraversalTranslateProjectFileItems)' != 'false' ">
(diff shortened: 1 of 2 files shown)


src/Traversal/Sdk/Sdk.targets
```diff
@@ -209,6 +213,49 @@
              SkipNonexistentTargets="$(SkipNonexistentTargets)" />
   </Target>
 
+  <Target Name="VSTest"
```
jeffkl: Could you please update the unit test that verifies targets?

src/Traversal/Sdk/Sdk.targets
```diff
@@ -59,6 +63,10 @@
     <PackDependsOn>
       ResolveReferences;
     </PackDependsOn>
+
+    <PublishDependsOn>
+      Build;
```
yhvicey: @jeffkl Could you please help review this? I'm not sure what should be put here, and just `Build` here seems works.

jeffkl: I don't believe these are needed.  The VSTest target doesn't really run in the traversal projects.  Instead I think it should just run the VSTest target for each project which already will build the project before running the target.  Although technically this will run the Build target in the traversal project which can run extra before/after stuff.  So maybe we do need it?  Did you test the behavior out locally and it did what you want?

yhvicey: I tried remove the Build target from VSTestDependsOn, but it seems failed to execute VSTest target on referenced projects. I roughly walk through the log but have no idea why it happens

yhvicey:
ping @jeffkl

src/Traversal/Sdk/Sdk.targets
jeffkl: I guess it shouldn't hurt, since publish depends on build for each project anyway.  We'll just get it out there and fix it if people report problems.

Thanks!

yhvicey:
It seems I must have some GPG-signed commit to bypass the branch policy... Is there any way to workaround this? or will I have to re-commit all the changes with GPG key?

jeffkl:
Sorry, I got slammed with work today and didn't merge.  I'll release a new package soon

yhvicey:
Thanks for your help!

PR #75, opened 2019-03-22
Set up CI with Azure Pipelines

rainersigwald:
Starting the process of moving the CI build to Azure Pipelines.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,5 @@
 pool:
-  name: VSEng-MicroBuildVS2019
-  demands: msbuild
+  name: Hosted VS2017
 
 variables:
   SignType: 'Test'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,6 +3,7 @@ pool:
 
 variables:
   SignType: 'Test'
+  Configuration: 'Debug'
 
 steps:
 - task: MSBuild@1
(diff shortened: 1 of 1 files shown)


jeffkl:
For PR builds, this is the [appveyor.yml](https://github.com/Microsoft/MSBuildSdks/blob/master/appveyor.yml)

The only artifacts should be the the logs in case of failure.  There's no need for the NUPKGs from a PR build.

MeikTranel:
> The only artifacts should be the the logs in case of failure.

Are we sure this is a good idea. As of right now this would publish binary logs, which contain basically everything. @KirillOsenkov taught me this is a big no-no. Builds happen on VSeng-MicroBuild.. are you guys sure there is no confidential environment variables?

jeffkl:
That's an interesting question.  Do public users have access to build artifacts?  If possible, we could upload the logs but restrict access to them.

azure-pipelines.yml
```diff
@@ -0,0 +1,27 @@
...
+variables:
+  SignType: 'Test'
+
+steps:
+- task: MSBuild@1
+  displayName: 'Build dirs.proj'
+  inputs:
+    solution: dirs.proj
+    platform: 'AnyCPU'
+    configuration: '$(Configuration)'
```
jeffkl: Just set this to Debug

rainersigwald:
Public users do have access to artifacts, but there should be no secrets available in the build environment. That's part of the reason we distinguish between `dnceng/public` and `dnceng/internal`.

azure-pipelines.yml
```diff
@@ -0,0 +1,27 @@
+pool:
+  name: Hosted VS2017
+
+variables:
+  SignType: 'Test'
```
jeffkl: Actually I like that this is test signing, because I've broken signing before with a PR.

MeikTranel:
@rainersigwald "dnceng/public and dnceng/internal." are these the azure devops orgs?

rainersigwald:
@MeikTranel Yes. Builds that run under https://dnceng.visualstudio.com/public/_build shouldn't have access to any secrets--they should be "as though you ran it on your machine" but automated. For builds that need to sign, push to feeds, etc, there's a separate infrastructure. This is (lightly) documented here: https://github.com/dotnet/arcade/blob/master/Documentation/AzureDevOps/AzureDevOpsOnboarding.md.

PR #88, opened 2019-04-26
Support static graph

cdmihai:


src/Traversal/Sdk/Sdk.targets
```diff
@@ -77,7 +88,7 @@
 
   <ItemGroup>
     <ProjectReference Update="@(ProjectReference)"
-                      Properties="$(TraversalGlobalProperties)" />
+                      AdditionalProperties="$(TraversalGlobalProperties)" />
```
cdmihai: Not sure if `TraversalGlobalProperties` is supposed to rewrite all properties, or just add themselves among the possibly other existing global properties

src/Traversal/Sdk/Sdk.targets
```diff
@@ -92,15 +103,17 @@
           DependsOnTargets="$(PrepareForBuildDependsOn)" />
 
   <Target Name="Build"
-          DependsOnTargets="$(BuildDependsOn)">
+          DependsOnTargets="$(BuildDependsOn)"
+          Condition=" '$(IsGraphBuild)' != 'true' " >
```
jeffkl: This seems funny to me that targets have to disable themselves when doing a graph build.  How come MSBuild runs anything at all?

cdmihai: This is due to a graph construction quirk (bug?) where references to `Microsoft.Net.Sdk` multitargeting projects from a non `Microsoft.Net.Sdk` project are not handled well. I first noticed this problem when looking into the traversal sdk. I plan to fix it eventually, but for now I went with the workaround of disabling running any targets in the traversal project.

PR #98, opened 2019-05-28
Allow loading .msbuildproj in VS using NoTargets

### LINKED ISSUE #97: Allow loading in VS for .msbuildproj using NoTargets SDK
It's quite common for `NoTargets` SDK projects to be just MSBuild scripts. Loading and editing these projects in VS would be useful for managing package references as well as editing and building from VS. 

Most of this support is already in place as long as the project extension is `.msbuildproj` (I'm using this in [AutoCodeFix](https://github.com/kzu/AutoCodeFix/blob/master/src/AutoCodeFix.Package/AutoCodeFix.Package.msbuildproj), for example). It would be nice if the two lines it takes to make it load in VS were included in the SDK, namely:

```xml
<Target Name="CompileDesignTime" />

<Import Project="$(MSBuildExtensionsPath)\Microsoft\VisualStudio\Managed\Microsoft.Managed.DesignTime.targets" Condition="Exists('$(MSBuildExtensionsPath)\Microsoft\VisualStudio\Managed\Microsoft.Managed.DesignTime.targets')" />
```

kzu:
By including the design-time targets and imports expected by CPS, we
allow the full loading and editing experience in the IDE automatically.

Fixes #97

Screenshot of this working in VS: 

image

### DIFF
--- a/src/NoTargets/Sdk/Sdk.props
+++ b/src/NoTargets/Sdk/Sdk.props
@@ -15,4 +15,8 @@
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" Condition=" '$(MicrosoftCommonPropsHasBeenImported)' != 'true' "/>
 
   <Import Project="$(CustomAfterNoTargetsProps)" Condition=" '$(CustomAfterNoTargetsProps)' != '' And Exists('$(CustomAfterNoTargetsProps)') " />
+
+  <!-- For CPS/VS support. Importing in .props allows any subsequent targets to redefine this if needed -->
+  <Target Name="CompileDesignTime" />
+  
 </Project>
--- a/src/NoTargets/Sdk/Sdk.targets
+++ b/src/NoTargets/Sdk/Sdk.targets
@@ -43,5 +43,10 @@
     </CoreBuildDependsOn>
   </PropertyGroup>
 
+  <!-- For CPS/VS support. See https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets#L60 -->
+  <Import Project="$(MSBuildExtensionsPath)\Microsoft\VisualStudio\Managed\Microsoft.Managed.DesignTime.targets" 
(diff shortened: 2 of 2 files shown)


kzu:
Heya @jeffkl, so, this is a low risk, high impact minor change, I think :). I'm using all over and it's pretty awesome, hehe.

jeffkl:
I am out on vacation today. I will review and test it tomorrow when I’m back.

src/NoTargets/Sdk/Sdk.props
```diff
@@ -10,9 +10,12 @@
     <UsingMicrosoftNoTargetsSdk>true</UsingMicrosoftNoTargetsSdk>
   </PropertyGroup>
 
+  <!-- For CPS/VS support. Importing in .props allows any subsequent targets to redefine this if needed -->
+  <Target Name="CompileDesignTime" />
```
jeffkl: Please move this to be after the `CustomBeforeNoTargetsProps` import

src/NoTargets/Sdk/Sdk.targets
```diff
@@ -44,4 +44,8 @@
   </PropertyGroup>
 
   <Import Project="$(CustomAfterNoTargets)" Condition="'$(CustomAfterNoTargets)' != '' and Exists('$(CustomAfterNoTargets)')" />
+  
+  <!-- For CPS/VS support. See https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/DesignTimeTargets/Microsoft.Managed.DesignTime.targets#L60 -->
```
jeffkl: Please move this to before the `CustomAfterNoTargets`

src/NoTargets/Sdk/Sdk.props
kzu: That would mean the target would not be overridable by that targets, say. Is that intentional?

src/NoTargets/Sdk/Sdk.targets
kzu: makes sense

src/NoTargets/Sdk/Sdk.props
kzu: Fixed anyway, it's not like targets are normally defined in props anyway...

PR #112, opened 2019-07-02
Add some usage info to README

drewnoakes:


README.md
```diff
@@ -29,7 +29,25 @@ Supports utility projects that do not compile an assembly.
 
 Supports staging artifacts from build outputs.
 
-## What Are MSBuild SDKS?
+## How can I use these SDKs?
+
+Start your project file with:
+
+```xml
+<Project Sdk="<package-name>/<version>">
```
jeffkl: I generally recommend people use `global.json` to specify the versions so that its only defined in one place.  That matters much more for Traversal and NoTargets and much less for CentralPackageVersions.  Should this explain the recommended way more to emphasize the use of global.json instead of the version in the project?  Or do you think that makes it harder to understand?

README.md
```diff
@@ -29,7 +29,25 @@ Supports utility projects that do not compile an assembly.
 
 Supports staging artifacts from build outputs.
 
-## What Are MSBuild SDKS?
+## How can I use these SDKs?
+
+Start your project file with:
```
jeffkl: Its probably worth really calling out here the fact that you **MUST** specify a version in order for the MSBuild project SDKs to be resolved.  I think that's the main point that is missing.  So whatever phrasing you think makes that clear would be good to add to this.

jeffkl:
@drewnoakes could you please address my comments? I'd like to merge this as it improves our documentation.

drewnoakes:
@jeffkl I've updated the PR so please take another look.

PR #117, opened 2019-07-11
Set IncludeBuildOutput to false in NoTargets

### LINKED ISSUE #116: NoTargets should set IncludeBuildOutput to false so that Pack task does not attempt to include an assembly
seems like dotnet pack expects an assembly in the output folder which is not generated of course:
 
error NU5026: Die im Paket einzubeziehende Datei "D:\Temp\_Demo\CrossCutting.Build\CrossCutting.Nunit\bin\Debug\netstandard1.1\CrossCutting.Nunit.dll

jeffkl:
This ensures the Pack task does not attempt to include build output from a NoTargets project since it won't emit an assembly.

Fixes #116

### DIFF
--- a/src/NoTargets.UnitTests/NoTargetsTests.cs
+++ b/src/NoTargets.UnitTests/NoTargetsTests.cs
@@ -39,6 +39,17 @@ namespace Microsoft.Build.NoTargets.UnitTests
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


jeffkl:
This is fixed in [Microsoft.Build.NoTargets 1.0.80](https://www.nuget.org/packages/Microsoft.Build.NoTargets/1.0.80)

Issue #101, opened 2019-06-07
[Microsoft.Build.CentralPackageVersions] Allow per project  package references that are not centrally managed.


amrmahdi:
**SDK: Microsoft.Build.CentralPackageVersions**
**Version: 2.0.36**

The current behavior of Microsoft.Build.CentralPackageVersions forces the consumer to completely opt-in and have all packages version centrally managed. 

It is desirable to have a subset of packages versions be centrally managed, and in the same time have per project packages that are not centrally managed. 

For example, one could have ``Microsoft.Extensions.*`` packages centrally managed in for all projects in a repo, and at the same time continue to reference packages per project as needed.  

A project would look like:

```xml
<PackageReference Include="Microsoft.Extensions.Configuration"  /> <!-- Centrally managed -->
<PackageReference Include="Grpc" Version="1.21.0" /> <!-- regular package reference -->
```

yhvicey:
I have no idea about how to distinguish updated `PackageReference` in `Packages.props` from other normal `PackageReference` without custom tasks, and I guess we have below options:

1. Add custom tasks to distinguish newly updated `PackageReference` in `Pacjages.props`;
    1. This may requires adding C# assembly to this SDK.
1. Allow user define versions in both places (`Packages.props` and project file) and only version in `Packages.props` will take effect; if user wants to override, they must use `VersionOverride` explicity;

Current behavior is:

1. All packages' version must be specified in `Packages.props`;
1. All project files cannot specify any version for any package;
1. If they want to do so, they must use `VersionOverride`.

If we choose option 1, behavior will be look like:

1. All packages' version must be specified at most once, no matter in `Packages.props` or project file;
1. Project files can have their own package versions, until it is also specified in `Packages.props`;
1. If it's already specified in `Packages.props` but need to be overrided, users can use `VersionOverride`.

If we choose option 2, behavior will be look like:

1. Packages can have their version specified in both `Packages.props` and project files;
1. Only version in `Packages.props` will take effect, and there's no error or warning if it's been specified twice;
1. User can override package version by using `VersionOverride`

All above descriptions based on one assumption: This SDK **CAN/WILLING TO** introduce such mechanism. @jeffkl any feedback?

jeffkl:
Sorry I was out on vacation.  There will have to be some convention of marking which PackageReferences are allowed to specify a version.  In your example above:

```xml
<PackageReference Include="Microsoft.Extensions.Configuration"  /> <!-- Centrally managed -->
[...]

PR #121, opened 2019-07-17
Support opt-out central package for sub projects

yhvicey:
Related to #101


yhvicey:
Current behavior:

1. All packages' version must be specified in `Packages.props`;
1. All project files cannot specify any version for any packages;
1. If they want to do so, they must use `VersionOverride`.

New behavior:

1. All packages' version must be specified at most once, no matter in `Packages.props` or project file;
1. Project files can have their own package versions, until it is also specified in `Packages.props`;
1. If it's already specified in `Packages.props` but need to be override, users can use `VersionOverride`.

yhvicey:
This implementation have some bug. I'm having some problem with how to distinguish `PackageReference` that specified their version twice (in `Packages.props` and project file). Will send an updated version later, including unit tests and fixes.

PR #139, opened 2019-11-15
Add support for static graph when project extension is not csproj, vbproj, or fsproj

cdmihai:


src/NoTargets/Sdk/Sdk.targets
```diff
@@ -70,5 +70,11 @@
   <Target Name="GetTargetPathWithTargetPlatformMoniker" />
 
   <Import Project="$(CustomAfterNoTargets)" Condition="'$(CustomAfterNoTargets)' != '' and Exists('$(CustomAfterNoTargets)')" />
-  
+
+  <!-- 
+    Microsoft.Managed.Targets is imported by the managed language target files, but most of the msbuild tasks are actually in Microsoft.Common.Currentversion.targets.
+    So import it when the managed targets do not get imported.
+  -->
+  <Import Project="$(MSBuildToolsPath)\Microsoft.Managed.targets" Condition="'$(ManagedLanguageTargetsGotImported)' != 'true'"/>
```
jeffkl: Maybe only do this if `LanguageTargets` is undefined?  That way you don't need the `ManagedLanguageTargetsGotImported` property which might change over time depending on file extensions in the SDK.  We should probably also add an escape hatch in case people need to disable this import
```xml
<!--
    If LangaugeTargets has a value, the SDK will import Microsoft.Common.targets but if the file has a different
    extension than .csproj or .vbproj then no targets are imported.
-->
<Import Project="$(MSBuildToolsPath)\Microsoft.Managed.targets" 
    Condition="'$(LanguageTargets)' == '' And '$(NoTargetsImportManagedTargets)' != 'false'" />
```

cdmihai: >Maybe only do this if LanguageTargets is undefined

Language targets is always defined. If no managed extension is detected, it falls back to `Microsoft.Common.targets`. Would you rather I verify whether `LanguageTargets` ends with `Microsoft.Common.targets`?

>We should probably also add an escape hatch in case people need to disable this import

The import is unconditional when the project file has the right extension, so the escape hatch would only turn off the imported I added here. Is that what you were thinking of as well?

jeffkl: I'm just worried about the hard coded list of extensions.  But I guess they won't change any time soon so its fine.

We can live without the escape hatch for now, if anyone hits an issues we'll add it later.

Issue #145, opened 2020-01-17
Artifacts potentially packs build scripts in wrong folder


chris-sekira:
The Artifacts project has props/targets that get packed into the _buildCrossTargeting_ folder. According to the NuGet package folder layout [documentation](https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#from-a-convention-based-working-directory) this should be the _buildMultiTargeting_ instead. 

Unless this some kind of a legacy workaround/disambiguation, the folder should probably be renamed. 

If I'm wrong on this and there's documentation that you've found for using the _buildCrossTargeting_, I'd be very appreciated if you could attach a link so that I can bookmark it for myself and request it's addition to the linked documentation above.

jeffkl:
NuGet renamed the folder and supports both at least for a while.  We should rename it.

https://github.com/NuGet/Home/issues/4098

chris-sekira:
Ah, google failed me. If _buildMultiTargeting_ is the new standard, I'll look into submitting a change. 

The oldest supported framework used here is net46, which was released before the linked issue's decision to move to _buildMultiTargeting_. Any chance we'd need to support _buildCrossTargeting_ for anyone using older clients?

jeffkl:
I can't think of a reason to keep both folders since people using older clients can just use an older package.  I think renaming the folder is the right solution.  Please feel free to contribute the change, otherwise I can get to it next week.

chris-sekira:
@jeffkl I'll make the change during lunch.

PR #146, opened 2020-01-17
Update Artifacts from buildCrossTargeting to buildMultiTargeting

chris-sekira:
Fix for #145

Moves build targets from legacy buildCrossTargeting folder to buildMultiTargeting


MeikTranel:
Not that this is very likely, but:
If you are using old CSProjs with this package, this might be a breaking change. Old csprojs hardcode import paths on first install. I'm not sure what the upgrade story looks like, but i'm pretty sure it doesnt migrate those imports. Just thought you might want to know.

jeffkl:
Thanks @MeikTranel, we'll see if anyone complains and if so add both folders.  I suspect that they can just use older versions of the package or update their stuff.

MeikTranel:
I'd prefer if they'd file an issue here: https://github.com/MeikTranel/MigrationSupport and i'll help them get off of old csprojs for good 😆

PR #153, opened 2020-02-25
Do not import Microsoft.Managed.targets in older versions of MSBuild.

### LINKED ISSUE #147: Microsoft.Build.NoTargets broken for MSBuild 15
Commit 38f7f902c291d41af120fd1640a2691bd329011d has "broken" `.msbuildproj` projects, as it does not correctly import the manged targets for `.msbuildproj` projects.

See:
https://github.com/microsoft/MSBuildSdks/blob/38f7f902c291d41af120fd1640a2691bd329011d/src/NoTargets/Sdk/Sdk.targets#L74-L78

This has resulted in errors like 
`C:\Users\NoOne\.nuget\packages\microsoft.build.notargets\1.0.88\Sdk\Sdk.targets(78,3): error MSB4019: The imported project "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\Microsoft.Managed.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk. [C:\B\A3\_work\16\s\src\SomeProject.msbuildproj]`

### Possible fix

As the `Microsoft.Managed.targets` file does not exist in earlier tool versions, and you only need it to support GraphBuild, one solution would be to import `Microsoft.Managed.targets` conditionally.

```xml
 <!-- 
    Microsoft.Managed.Targets is imported by the managed language target files, but most of the msbuild tasks are actually in Microsoft.Common.Currentversion.targets.
    So import it when the managed targets do not get imported, to support GraphBuild
  -->
  <Import Project="$(MSBuildToolsPath)\Microsoft.Managed.targets" Condition="'$(ManagedLanguageTargetsGotImported)' != 'true' and Exists('$(MSBuildToolsPath)\Microsoft.Managed.targets')"/>
```

jeffkl:
Microsoft.Managed.targets was added in MSBuild 16 so don't import it in MSBuild 15 and below

Fixes #147

### DIFF
--- a/src/NoTargets/Sdk/Sdk.targets
+++ b/src/NoTargets/Sdk/Sdk.targets
@@ -72,9 +72,9 @@
   <Import Project="$(CustomAfterNoTargets)" Condition="'$(CustomAfterNoTargets)' != '' and Exists('$(CustomAfterNoTargets)')" />
 
   <!-- 
-    Microsoft.Managed.Targets is imported by the managed language target files, but most of the msbuild tasks are actually in Microsoft.Common.Currentversion.targets.
+    Microsoft.Managed.Targets is imported by the managed language target files in MSBuild 16.0 and above, but most of the msbuild tasks are actually in Microsoft.Common.Currentversion.targets.
     So import it when the managed targets do not get imported.
   -->
-  <Import Project="$(MSBuildToolsPath)\Microsoft.Managed.targets" Condition="'$(ManagedLanguageTargetsGotImported)' != 'true'"/>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.Managed.targets" Condition="'$(MSBuildAssemblyVersion)' >= '16.0' And '$(ManagedLanguageTargetsGotImported)' != 'true'" />
 
 </Project>


src/NoTargets/Sdk/Sdk.targets
```diff
@@ -72,9 +72,9 @@
   <Import Project="$(CustomAfterNoTargets)" Condition="'$(CustomAfterNoTargets)' != '' and Exists('$(CustomAfterNoTargets)')" />
 
   <!-- 
-    Microsoft.Managed.Targets is imported by the managed language target files, but most of the msbuild tasks are actually in Microsoft.Common.Currentversion.targets.
+    Microsoft.Managed.Targets is imported by the managed language target files in MSBuild 16.0 and above, but most of the msbuild tasks are actually in Microsoft.Common.Currentversion.targets.
     So import it when the managed targets do not get imported.
   -->
-  <Import Project="$(MSBuildToolsPath)\Microsoft.Managed.targets" Condition="'$(ManagedLanguageTargetsGotImported)' != 'true'"/>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.Managed.targets" Condition="'$(MSBuildAssemblyVersion)' >= '16.0' And '$(ManagedLanguageTargetsGotImported)' != 'true'" />
```
KalleOlaviNiemitalo: I was told in <https://github.com/MicrosoftDocs/visualstudio-docs/issues/4696#issuecomment-586433557> that MSBuildAssemblyVersion is superseded by MSBuildVersion. Is MSBuildVersion not yet available in some of the versions you want to recognize here?

jeffkl: Yeah technically `MSBuildVersion` is the `Major.Minor.Build` and `MSBuildAssemblyVersion` is just `Major.0`

https://github.com/microsoft/msbuild/blob/master/src/Build/Evaluation/Evaluator.cs#L1138-L1139

`MSBuildAssemblyVersion` won't be going away and `MSBuildVersion` is more precise since Microsoft started shipping more point releases like 16.1, 16.2, etc we needed more precision.  ([I actually added](https://github.com/microsoft/msbuild/pull/2819) `MSBuildVersion`)

Technically in this case I only care about MSBuild 15 vs MSBuild 16 so `MSBuildAssemblyVersion` will work.

PR #157, opened 2020-03-24
Add BuildInParallel and ContinueOnError ext points

ViktorHofer:
In dotnet/runtime we need more fine grained control over the `BuildInParallel` and `ContinueOnError` setting for the `Test` target. For the sake of consistency I added extension points for the other targets as well. 

For an example where in dotnet/runtime this is needed, see https://github.com/dotnet/runtime/blob/master/src/libraries/tests.proj.

cc @jeffkl

### DIFF
--- a/src/Traversal/Sdk/Sdk.targets
+++ b/src/Traversal/Sdk/Sdk.targets
@@ -104,7 +104,7 @@
 
   <PropertyGroup>
     <BuildInParallel Condition="'$(BuildInParallel)' == ''">true</BuildInParallel>
-    <ContinueOnErrorCondition="'$(ContinueOnError)' == ''">false</ContinueOnError>
+    <ContinueOnError Condition="'$(ContinueOnError)' == ''">false</ContinueOnError>
   </PropertyGroup>
 
   <Target Name="PrepareForBuild"
--- a/src/Traversal/Sdk/Sdk.targets
+++ b/src/Traversal/Sdk/Sdk.targets
@@ -102,6 +102,11 @@
     <ProjectReference Remove="$(MSBuildProjectFullPath)" />
   </ItemGroup>
 
+  <PropertyGroup>
+    <BuildInParallel Condition="'$(BuildInParallel)' == ''">true</BuildInParallel>
+    <ContinueOnErrorCondition="'$(ContinueOnError)' == ''">false</ContinueOnError>
+  </PropertyGroup>
+
   <Target Name="PrepareForBuild"
           DependsOnTargets="$(PrepareForBuildDependsOn)" />
 
@@ -120,10 +125,10 @@
           Condition=" '$(IsGraphBuild)' != 'true' ">
     <MSBuild Projects="@(ProjectReference)"
              Targets="Clean"
-             BuildInParallel="$([MSBuild]::ValueOrDefault('$(CleanInParallel)', '$(BuildInParallel)')"
+             BuildInParallel="$([MSBuild]::ValueOrDefault('$(CleanInParallel)', '$(BuildInParallel)'))"
              SkipNonexistentProjects="$(SkipNonexistentProjects)"
              SkipNonexistentTargets="$(SkipNonexistentTargets)"
-             ContinueOnError="$([MSBuild]::ValueOrDefault('$(CleanContinueOnError)', '$(ContinueOnError)')" />
+             ContinueOnError="$([MSBuild]::ValueOrDefault('$(CleanContinueOnError)', '$(ContinueOnError)'))" />
   </Target>
 
   <Target Name="Test"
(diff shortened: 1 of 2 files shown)


src/Traversal/Sdk/Sdk.targets
```diff
@@ -111,47 +126,67 @@
...
+    <Error Condition="'$(ContinueOnError)' == 'ErrorAndContinue' and '$(MSBuildLastTaskResult)' == 'false'" />
   </Target>
 
   <Target Name="Clean"
           DependsOnTargets="$(CleanDependsOn)"
           Condition=" '$(IsGraphBuild)' != 'true' ">
     <MSBuild Projects="@(ProjectReference)"
              Targets="Clean"
-             BuildInParallel="$(BuildInParallel)"
+             BuildInParallel="$(CleanInParallel)"
```
jeffkl: I have this aversion for creating properties which can be defaulted in line since its just more strings.  Can you change this to:
```xml
  BuildInParallel="$([MSBuild]::ValueOrDefault('$(CleanInParallel)', '$(BuildInParallel)')"
```
That's effectively the same thing and lets you get rid of all the new properties being added if you use this pattern.  Logs still show what value is used so its not any harder to debug.

src/Traversal/Sdk/Sdk.targets
```diff
@@ -111,47 +126,67 @@
     <MSBuild Projects="@(ProjectReference)"
              BuildInParallel="$(BuildInParallel)"
              SkipNonexistentProjects="$(SkipNonexistentProjects)"
-             SkipNonexistentTargets="$(SkipNonexistentTargets)" />
+             SkipNonexistentTargets="$(SkipNonexistentTargets)"
+             ContinueOnError="$(ContinueOnError)" />
+    
+    <!-- If we ErrorAndContinue we need to propagate the error. -->
+    <Error Condition="'$(ContinueOnError)' == 'ErrorAndContinue' and '$(MSBuildLastTaskResult)' == 'false'" />
```
jeffkl: I have never seen this pattern before.  Doesn't ErrorAndContinue mean to still treat the task as an error but then continue executing tasks in the target?  But then we call `<Error />` with no message in that case? I'm assuming there's a logical explanation, can you fill me in? 😄

ViktorHofer: I noticed that in dotnet/runtime, I think that came from here: https://github.com/dotnet/corefx/commit/60967f18fd7b47dff0576759e3ea6dcb1ea3cbcb

ViktorHofer: ok as discussed offline, unnecessary.

src/Traversal/Sdk/Sdk.targets
ViktorHofer: done :) thanks for sharing the pattern. that's super helpful.

src/Traversal/Sdk/Sdk.targets
```diff
@@ -111,47 +111,52 @@
...
+             ContinueOnError="$([MSBuild]::ValueOrDefault('$(CleanContinueOnError)', '$(ContinueOnError)')" />
   </Target>
 
   <Target Name="Test"
           DependsOnTargets="$(TestDependsOn)"
           Condition=" '$(IsGraphBuild)' != 'true' ">
     <MSBuild Projects="@(ProjectReference)"
              Targets="Test"
-             BuildInParallel="$(BuildInParallel)"
+             BuildInParallel="$([MSBuild]::ValueOrDefault('$(TestInParallel)', '$(BuildInParallel)')"
```
jeffkl: You're missing a closing parenthesis

```diff
-BuildInParallel="$([MSBuild]::ValueOrDefault('$(TestInParallel)', '$(BuildInParallel)')"
+BuildInParallel="$([MSBuild]::ValueOrDefault('$(TestInParallel)', '$(BuildInParallel)'))"
```

And it turns out that this gets parsed down to an empty string and MSBuild still doesn't like that.

I think we'll have to default `BuildInParallel` and `ContinueOnError`:

```xml
<PropertyGroup>
  <BuildInParallel Condition="'$(BuildInParallel)' == ''">true</BuildInParallel>
  <ContinueOnErrorCondition="'$(ContinueOnError)' == ''">false</ContinueOnError>
</PropertyGroup>
```

Or we'll have to use a more complex syntax:
```xml
ContinueOnError="$([MSBuild]::ValueOrDefault($([MSBuild]::ValueOrDefault('$(CleanContinueOnError)', '$(ContinueOnError)')), 'false'))"
```

What do you think is better?

ViktorHofer: Oh sorry, lol. 

> What do you think is better?

I believe we shouldn't set defaults here in the SDK... just my 5c. What if I revert this to the state before which requires the hardcoded props but still better than defaulting something here?

jeffkl: These are already set in official SDKs.

https://github.com/microsoft/msbuild/blob/64c8e4a18e7cf7f064fbad304ea7ed877cdaa0a1/src/Tasks/Microsoft.Common.CurrentVersion.targets#L342
https://github.com/microsoft/msbuild/blob/64c8e4a18e7cf7f064fbad304ea7ed877cdaa0a1/src/Tasks/Microsoft.Common.CurrentVersion.targets#L610

So I'm comfortable setting them in Traversal as well.

PR #158, opened 2020-03-27
Filter projects per target and add BuildInParallel extension point

ViktorHofer:
Adding an extension point to filter projcts per target via attributes
and allow to batch on BuildInParallel.

cc @jeffkl

### DIFF
--- a/src/Traversal/README.md
+++ b/src/Traversal/README.md
@@ -108,3 +108,23 @@ Change the `BuildInParallel` setting for the Test target.
   </PropertyGroup>
 </Project>
 ```
+
+The following attributes can be set to false to exclude ProjectReferences for a specific target:
+- Build
+- Clean
+- Test
+- Pack
+- Publish
+
+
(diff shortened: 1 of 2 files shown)


src/Traversal/README.md
```diff
@@ -108,3 +108,23 @@ Change the `BuildInParallel` setting for the Test target.
...
+
+
+**Example**
+
+Add the `Test` attribute to the `ProjectReference` to exclude it when invoking the Test target.
+
+```xml
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <PackageReference Include="ProjectA.csproj" Test="false" />
```
CZEMacLeod: Shouldn't this be `ProjectReference` rather than `PackageReference`

ViktorHofer: you are right, thanks will send a follow-up PR to fix that.

PR #160, opened 2020-03-30
Fix AdditionalProperties overwrite in Traversal

ViktorHofer:
AdditionalProperties are currently always overwritten in ProjectReferences. Fixing that by adding to instead of setting the property.

@jeffkl can we please publish another release with this fix. Last time I'll nag you for this :)

### DIFF
--- a/src/Traversal/Sdk/Sdk.targets
+++ b/src/Traversal/Sdk/Sdk.targets
@@ -89,9 +89,9 @@
     <ProjectFile Remove="@(ProjectFile)" />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TraversalGlobalProperties)' != ''">
     <ProjectReference Update="@(ProjectReference)"
-                      AdditionalProperties="$(TraversalGlobalProperties)" />
+                      AdditionalProperties="%(AdditionalProperties);$(TraversalGlobalProperties)" />
   </ItemGroup>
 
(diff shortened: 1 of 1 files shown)


src/Traversal/Sdk/Sdk.targets
```diff
@@ -89,9 +89,9 @@
     <ProjectFile Remove="@(ProjectFile)" />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TraversalGlobalProperties)' != ''">
     <ProjectReference Update="@(ProjectReference)"
-                      AdditionalProperties="$(TraversalGlobalProperties)" />
+                      AdditionalProperties="%(AdditionalProperties);$(TraversalGlobalProperties)" />
```
jeffkl: Do you have time to write a unit test?

ViktorHofer: I just looked into the tests for Traversal and it would probably take me an hour or more to familiarize myself with msbuild testing. I filed an issue to track this https://github.com/microsoft/MSBuildSdks/issues/161 and I hope we can get this merged meanwhile to unblock my work in dotnet/runtime and I'll hopefully have time to follow-up on the issue EOW. Does that sound good @jeffkl?

jeffkl: Sounds good.  I'll add the test, thanks for filing the tracking issue.

Issue #178, opened 2020-05-19
Disallow VersionOverrides without a central version


erikness:
Right now, if you make a PackageReference with no version and no entry in Packages.props, you'll get an error. But this SDK does not actually force every PackageReference to have an entry in Packages.props: if the PackageReference uses a VersionOverride, it can be missing from Packages.props.

It's not a conceptual problem - no need to check Packages.props when you already know the exact version you want, right? - but nevertheless I think it can be a hygiene problem. So I'd like to propose throwing errors for these as well.

## Reasons

First, some teams probably use Packages.props as a big list of NuGet packages that are used in their repo. If there are PackageReferences that are not recorded in Packages.props, you lose the easy ability to see your entire group of package dependencies. (Though, on the other hand, it's also possible for you to have a Packages.props global version without any project actually using that package, so maybe teams shouldn't assume this).

Second, VersionOverride is an awkward name if it's not actually overriding anything.

Third, I don't see an easy way to implement this restriction for just my team; if it were easy to make a target that did this, it would be one thing, but the only implementation I can think of involves editing Sdk.targets. So it would be helpful if this feature were provided by the SDK itself. (Alternative: maybe add a new extensibility property, CustomAfterCentralPackagesImport?)

This feature would be team-specific (I'm just going off my own experiences and needs- it's not like I've done a user study) and it would be a breaking change (though not in a particularly disruptive way) so of course it can be behind a feature flag.

## Sketch of implementation

In Sdk.targets, right after importing Packages.props, you have a bunch of PackageReference items with both VersionOverride and Version metadata. But you also might have PackageReference items with just VersionOverride metadata and no Version - these are the "illegal" ones with respect to this feature.

```
<_PackageReferenceWithVersionOverrideButNoCentralVersion 
    Include="@(PackageReference->HasMetadata('VersionOverride'))"
    Exclude="@(PackageReference->HasMetadata('Version'))"
 />   
```

Then, you'd just place an <Error> task at the end of the CheckPackageReferences target if there are any of these items. I don't think you'd need to change any other behavior of CentralPackageVersions.

jeffkl:
Sounds like a good feature, do you want to send a pull request?  If not, I can work on it next week as I'm on vacation at the moment.

erikness:
I'll make a pull request in the next few days. Though I think I found a good way to do this in my own team's repo, and this might be a reason to reject the PR:

[...]

PR #180, opened 2020-05-27
Disallow orphaned VersionOverrides (opt-in)

erikness:
See #178 for a description of the problem.

It's a small diff, but it's my first time contributing to this repo so I have some notes and concerns, listed in a comment below. I also have a way you could implement this in a code base with including this feature in the SDK; this should help us understand whether this PR is worth merging.


erikness:
Notes and concerns:

- My item/property names are verbose, I'd really appreciate more concise ideas
- Does DisableVersionOverridesWithoutCentralVersions clearly indicate that by turning this on you're disallowing a certain kind of <PackageReference>? ("turn this on in order to turn this other thing off" can be confusing).
- DisableVersionOverridesWithoutCentralVersions must be 'true' for this feature to run, and the SDK doesn't set this anywhere, so by default it's the empty string and thus treated as false. This seems OK to me but it's inconsistent with other properties, empty string is true because we check for `!= 'false'`.
- Like the other Errors, these new Errors are raised at evaluation time, so they are raised early and apply to any target (Restore, Build, etc.)
- For _PackageReferenceWithVersionOverrideButNoCentralVersion, I used an Include/Exclude pattern instead of a Condition because MSBuild doesn't allow me to batch on custom metadata here (I don't know why)
- This feature is turned off when EnablePackageVersionOverride is false (even if DisableVersionOverridesWithoutCentralVersions is true)
- This feature is turned off when EnableCentralPackageVersions is false (even if DisableVersionOverridesWithoutCentralVersions is true)
- I didn't do anything about implicitly defined packages so I think the feature should apply to them too. Should I handle these specially? Do implicitly defined packages like FSharp.Core use VersionOverride or just Version? I don't know the pattern here.
- I didn't do anything about GlobalPackageReference either. My items are collected right after GlobalPackageReferences are copied to PackageReferences (an arbitrary choice). But GlobalPackageReferences are not allowed to have VersionOverrides (only Versions) so this feature would never apply to them
    - In principle, I think you *can* do a VersionOverride instead of Version for your GlobalPackageReference and it'll have the same effect (even though it's wrong). However, with this feature turned on, specifying VersionOverride will cause the new Error to be raised. It's kind of an indirect error message but I guess it's ok.
- Currently, this gives you one error at a time - so if there are multiple violations in a project, you'd have to build -> fix that one PackageReference -> build again -> fix the next PackageReference -> etc. We could, alternatively, use `ContinueOnError="ErrorAndContinue"` on the <Error /> task, to spit out all the violations at once - but then the build wouldn't stop here. I prefer this alternative personally, but other <Error /> task invocations above, don't do this, so I kept my new invocaton in line with the existing invocations for now.
- This doesn't have an error code so projects can't suppress it. I think this is probably fine. If developers want to have exceptions to this rule, then they should just keep this feature turned off.
- Testing:
    - I consumed this in a local repo of mine and made sure it raised errors on "bad" PackageReferences
    - 2 basic unit tests, verifying that orphaned VersionOverrides are allowed normally and disallowed when you turn this feature on

erikness:
How to implement this yourself in a repo:

First, put this at the end of Packages.props, after all the version lines:

```
  <ItemGroup>
    <PackageReferenceWithoutGlobalVersion Include="@(PackageReference->HasMetadata('VersionOverride'))"
      Exclude="@(PackageReference->HasMetadata('Version'))" />
  </ItemGroup>
```

Second, put this target in a global build file:

```
  <Target Name="ErrorOnPackageReferencesWithoutGlobalVersion" Inputs="@(PackageReferenceWithoutGlobalVersion)"
    Outputs="%(Identity).unused" BeforeTargets="BeforeBuild">
    <Error Text="The PackageReference to %(PackageReferenceWithoutGlobalVersion.Identity) needs an entry in Packages.props"
      ContinueOnError="ErrorAndContinue" />
  </Target>
```

jeffkl:
Sorry I was out of town last week.  I'm looking at this now.  The unit tests weren't running as part of the build because of something else which I've fixed via a commit.  So you'll need to rebase and push when you get a chance
```
git fetch upstream master
git rebase upstream/master
git push
```

If that doesn't work, I can update the branch in the UI here and you can just pull.

src/CentralPackageVersions/README.md
```diff
@@ -59,6 +59,8 @@ In some cases, you may need to override the version for a particular project.  T
 </Project>
 ```
 
+A `PackageReference` does not have to have a corresponding central version if it has a `VersionOverride`. If you want to require a central version for each `PackageReference` in your code base, set `DisableVersionOverridesWithoutCentralVersions` to `true`.
```
jeffkl, line 62: The other control mechanism is named `EnablePackageVersionOverride` which defaults to `true`. 

I think we should name this new property `EnablePackageVersionOverrideWithoutCentralVersion` which also defauls to `true` and you set it to `false` to generate the error.  Seems like keeping it consistent will give a better user experience.

src/CentralPackageVersions/Sdk/Sdk.targets
```diff
@@ -28,6 +28,8 @@
     <EnableCentralPackageVersions
       Condition="'$(EnableCentralPackageVersions)' == '' And (!Exists('$(CentralPackagesFile)') Or Exists('$(MSBuildProjectDirectory)\packages.config') Or '$(MSBuildProjectExtension)' == '.vcxproj' Or '$(MSBuildProjectExtension)' == '.ccproj' Or '$(MSBuildProjectExtension)' == '.nuproj')">false</EnableCentralPackageVersions>
 
+    <EnablePackageVersionOverrideWithoutCentralVersion Condition="'$(EnablePackageVersionOverrideWithoutCentralVersion)' == '' And '$(EnableCentralPackageVersions)' != 'false'">true</EnablePackageVersionOverrideWithoutCentralVersion>
```
jeffkl, line 31: There's no need to default this property to any value as we'll only look for the user to turn it off by setting it to `false`.  Let's remove this.

erikness: Yeah I wasn't sure whether checking for false and setting true by default was the right choice. I removed this, thanks for the guidance here.

PR #183, opened 2020-05-28
Collect and bubble up "Build" TargetOutputs

### LINKED ISSUE #181: TraversalSDK: Collect TargetOutputs of ProjectReference
Similar to how the sln metaproj produces a collective TargetOutput of all projects built, the traversal SDK should list those when building as well.

**Expectation:**
Running
`msbuild C:/traversaltest/traverse.proj /t:Build`
```xml
<Project Sdk="Microsoft.Build.Traversal/2.0.34">
  <ItemGroup>
    <ProjectReference Include="a/a.csproj" />
    <ProjectReference Include="b/b.csproj" />
  </ItemGroup>
</Project>
```
should list:

C:/traversaltest/a/bin/debug/netcoreapp3.1/a.dll,
C:/traversaltest/b/bin/debug/netcoreapp3.1/b.dll

as TargetOutputs of target "Build".

While "Build" is my primary concern respective targetoutputs of other targets should certainly work the same way they do in the regular SDK.
_When testing this make sure to verify SDK behavior when using x-targeted projects because those might have some issues producing outputs from inner builds themselves._

MeikTranel:
Resolves #181

### DIFF
--- a/src/Traversal.UnitTests/TraversalTests.cs
+++ b/src/Traversal.UnitTests/TraversalTests.cs
@@ -302,5 +302,48 @@ namespace Microsoft.Build.Traversal.UnitTests
 
             result.ShouldBeTrue(customMessage: () => buildOutput.GetConsoleLog());
         }
+
+        [Fact]
+        public void CollectsProjectReferenceBuildTargetOutputs()
+        {
+            string[] projects = new[]
+            {
+                GetSkeletonCSProjWithTargetOutputs("A"),
+                GetSkeletonCSProjWithTargetOutputs("B")
+            }.Select(i => i.FullPath).ToArray();
+
+            var subTraversalProject = ProjectCreator
+                .Templates
+                .TraversalProject(projects, path: GetTempFile("dirs.proj"))
+                .Save();
+
+            ProjectCreator
+                .Create(path: GetTempFile("root.proj"))
+                .Target("BuildTraversalProject")
+                .Task(
+                    "MSBuild",
+                    parameters: new Dictionary<string, string>
+                    {
+                        ["Projects"] = subTraversalProject.FullPath,
+                        ["Targets"] = "Build"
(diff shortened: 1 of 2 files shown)


MeikTranel:
image

This is how it looks when calling a traversal project now. I couldn't bear seeing another level of collectedbuildoutput as the variable name 😄

MeikTranel:
Don't know if this is intended but the testing stage of the AzDO pipeline doesn't seem to run any tests.

jeffkl:
Thanks @MeikTranel, I've fixed the unit test execution.  Do you want to rebase and push or should I update your branch in the UI here?

MeikTranel:
Should be good now.

MeikTranel:
No Problem. Is it possible to get this released to nuget so i can use it? Working on something right now where this would be real handy 🤣

jeffkl:
Yeah sorry was working on some signing issues.  A new version is on its way to NuGet.

https://www.nuget.org/packages/Microsoft.Build.Traversal

MeikTranel:
Haha absolutely no problemo

ViktorHofer:
@MeikTranel great contributions, thanks! Unfortunately this doesn't work for multi-targeting projects: https://github.com/microsoft/msbuild/issues/2148.

MeikTranel:
Yeah i was aware of this when i did this PR. One can work around it, but ultimately the dotnet sdk should do this.
Its not really an issue with the traversal sdk because we have no concept of TFM resolution for projectreference and thus depend upon the dependency being as autonomous as possible. One day the publish scenario should be looked at, but thats alot of work.
Also: I think if you attach a TFM via AdditionalProperties to the Reference you get the regular Outputs back, because it doesnt override the build task, because its consider InnerLoop already when it enters the Multi-TFM-Project build target.

PR #190, opened 2020-06-14
Fix bubbling of TargetOutputs for "Rebuild" calls

MeikTranel:
I took the freedom to pin down the .NET SDK version via global.json - i used the same version specified in the AzDO pipeline.
.NET 5 Previews were failing builds with this codebase on my machine:
image
Further explanation in the respective commit message.

I hope the additional change is cool with you ;)

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,4 +1,8 @@
 {
+  "sdk": {
+    "version": "3.1.300",
+    "rollForward": "latestFeature"
+  },
   "msbuild-sdks": {
     "Microsoft.Build.NoTargets": "1.0.94"
   }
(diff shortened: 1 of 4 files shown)


global.json
```diff
@@ -1,4 +1,8 @@
 {
+  "sdk": {
```
jeffkl, line 2: This is fine, but lets not have the version in two places.  Remove the version from the pipeline YAML and update the UseDotNet task like this:

```diff
variables:
  ArtifactsDirectoryName: 'artifacts'
  BinaryLoggerArgs: '"/BinaryLogger:$(Build.SourcesDirectory)\$(ArtifactsDirectoryName)\msbuild.binlog"'
  BuildConfiguration: 'Debug'
  BuildPlatform: 'Any CPU'
-  DotNetCoreVersion: '3.1.300'
...
  - task: UseDotNet@2
-    displayName: 'Install .NET Core $(DotNetCoreVersion)'
+    displayName: 'Install .NET Core from global.json'
    inputs:
-      version: '$(DotNetCoreVersion)'
+      useGlobalJson: true
```

MeikTranel: I didn't know this works. Hell yeah. Give me a sec.

PR #192, opened 2020-07-05
Capture cross-targeting build TargetOutputs

ViktorHofer:
With https://github.com/microsoft/MSBuildSdks/commit/9dca0b7884bde58d16eff4a41359bf36230bd368, capturing the build output is now possible but `CollectedBuildOutput` does not include output from cross-targeting projects. Overriding the Common target's Build output to allow that.

cc @MeikTranel @jeffkl


ViktorHofer:
Closing as this would need to be imported by the child projects, not the traversal one.

MeikTranel:
You could force the crosstargeted project from the traversal project to import certain targets, but that would be so bushfix-esque that it should probably be done by the individual user not in the package. Ultimately I think the upstream issue in the dotnet sdk should be reconsidered.

PR #197, opened 2020-08-25
Fix restore in Traversal when TargetFrameworks property is set

### LINKED ISSUE #196: Restoring traversal project fails in .NET 5 SDK if TargetFrameworks specified
Some breaking changes in the .NET SDK mean that a previously innocuous property, `TargetFrameworks`, will cause restore to fail.  This is because the property is used to generate two entries in the restore graph but some logic that translates TargetFramework to the expanded TargetFrameworkMoniker property is not imported by Microsoft.Build.Traversal.

We'll need to empty the `TargetFrameworks` property so that NuGet restore doesn't think the project is multi-targeting since Traversal projects don't support multi-targeting anyway.

Repro:

```xml
<Project Sdk="Microsoft.Build.Traversal/2.0.24">
  <PropertyGroup>
    <TargetFrameworks>net45;netcoreapp3.1</TargetFrameworks>
  </PropertyGroup>
  <ItemGroup>
    <ProjectReference Include="ProjectA\ProjectA.csproj" />
    <ProjectReference Include="ProjectB\ProjectB.csproj" />
  </ItemGroup>
</Project>
```

```cmd
msbuild /t:Restore

Microsoft (R) Build Engine version 16.8.0-preview-20411-05+9a32a8063 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
C:\Program Files (x86)\Visual Studio 2019 Preview\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets(128,5): error : Invalid restore input. Duplicate frameworks found: 'net40, net40'. Input files: \dirs.proj. [dirs.proj]
Build FAILED.

C:\Program Files (x86)\Visual Studio 2019 Preview\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets(128,5): error : Invalid restore input. Duplicate frameworks found: 'net40, net40'. Input files: \dirs.proj. [dirs.proj]
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:05.72
```

jeffkl:
A breaking change in the .NET SDK now uses TargetFrameworks differently leading to restore errors when Traversal projects set it accidentally.

Traversal projects do not support multi-targeting by design, so this change just sets TargetFrameworks to empty to ensure that NuGet does not try to restore using multiple frameworks.

Fixes #196

### DIFF
--- a/src/Traversal/Sdk/Sdk.targets
+++ b/src/Traversal/Sdk/Sdk.targets
@@ -18,6 +18,15 @@
     <OutputPath Condition=" '$(Configuration)' != '' And '$(Platform)' != '' ">bin\$(Configuration)\$(Platform)\</OutputPath>
   </PropertyGroup>
 
+  <!--
+    TargetFrameworks is used to convey that a project targets multiple frameworks.  Traversal does not support multi-targeting
+    but having this property set to a value can confuse NuGet leading to errors.  Setting the property to empty fixes an issue
+    where a user unknowningly sets it, breaking restore.  See also: https://github.com/microsoft/MSBuildSdks/issues/196
+  -->
+  <PropertyGroup>
+    <TargetFrameworks></TargetFrameworks>
+  </PropertyGroup>
+
   <Import Project="$(MSBuildToolsPath)\Microsoft.Common.targets" Condition=" Exists('$(MSBuildToolsPath)\Microsoft.Common.targets') " />
 
   <ItemGroup>
--- a/src/Traversal.UnitTests/TraversalTests.cs
+++ b/src/Traversal.UnitTests/TraversalTests.cs
@@ -9,6 +9,7 @@ using Microsoft.Build.Utilities.ProjectCreation;
 using Shouldly;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using UnitTest.Common;
 using Xunit;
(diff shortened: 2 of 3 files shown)


src/Traversal/Sdk/Sdk.targets
```diff
@@ -18,6 +18,15 @@
     <OutputPath Condition=" '$(Configuration)' != '' And '$(Platform)' != '' ">bin\$(Configuration)\$(Platform)\</OutputPath>
   </PropertyGroup>
 
+  <!--
+    TargetFrameworks is used to convey that a project targets multiple frameworks.  Traversal does not support multi-targeting
```
tmds, line 22: >  Traversal does not support multi-targeting

@jeffkl what does this mean? Can the `ProjectReference` projects can still use `TargetFrameworks` with multiple values?

jeffkl: Traversal projects don't really support a target framework, since they aren't built into an assembly that is bound to a particular framework.  This means that things like NuGet do not need to confirm that the target framework is compatible, the traversal project is just basically telling the system to build a set of projects.  The ProjectReference in an actual project does need to run through a compatibility check since it will be a loadable assembly though.

But ProjectReference items in traversal aren't the same, we just use the same item for consistency.  But yes, actual projects still get to use multiple target frameworks.

PR #203, opened 2020-09-02
Allow per-project PackageReference

yhvicey:
Impl #101.

### DIFF
--- a/src/CentralPackageVersions/Sdk/Sdk.targets
+++ b/src/CentralPackageVersions/Sdk/Sdk.targets
@@ -29,6 +29,12 @@
       Condition="'$(EnableCentralPackageVersions)' == '' And (!Exists('$(CentralPackagesFile)') Or Exists('$(MSBuildProjectDirectory)\packages.config') Or '$(MSBuildProjectExtension)' == '.vcxproj' Or '$(MSBuildProjectExtension)' == '.ccproj' Or '$(MSBuildProjectExtension)' == '.nuproj')">false</EnableCentralPackageVersions>
 
     <!--
+      Allow users to specify PacakgeReference with Version in project file without showing up in CentralPackagesFile
+    -->
+    <AllowPackageReferenceWithVersion
+      Condition=" '$(AllowPackageReferenceWithVersion)' == '' ">false</AllowPackageReferenceWithVersion>
+
+    <!--
       Include this file and the Packages.props file (if necessary) in MSBuildAllProjects so that rebuilds happen if the Packages.props changes.
     -->
     <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)' != 'Current'">$(MSBuildAllProjects);$(MsBuildThisFileFullPath)</MSBuildAllProjects>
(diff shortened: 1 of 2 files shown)


src/CentralPackageVersions/Sdk/Sdk.targets
```diff
@@ -191,6 +207,18 @@
       File="$(MSBuildProjectFullPath)" />
   </Target>
 
+  <Target Name="RestoreOriginalPackageReferenceWithVersion"
```
jeffkl, line 210: This won't work for legacy CSPROJ projects (ones that don't use Microsoft.NET.Sdk) because in Visual Studio it does not run targets before restoring.  This means you must create the list in a top-level `<ItemGroup />` and cannot manipulate anything in a target.

yhvicey: Not sure if duplicate declaration will lead into any problem.

- Project declared `PackageReference` with `Version`, but not declared in `Packages.props`
  This should be ok
- Project declared `PackageReference` with `Version`, and also declared in `Packages.props`
  `PackageReference`s in `Packages.props` should override version specified in project file, I guess
- Project declared `PackageReference` with `VersionOverride`, but not declared in `Packages.props`
  This should cause error
- Project declared `PackageReference` with `VersionOverride`, and also declared in `Packages.props`
  This is normal case

jeffkl: Yes those cases look correct.  Here's an example of how to do it all in the top-level item groups and avoid a target:
```xml
<Project>
  <!-- Pretend this is declared in the project -->
  <ItemGroup>
    <PackageReference Include="Foo" Version="1.0.0" />
    <PackageReference Include="Bar" />
  </ItemGroup>

  <!-- This is declared in Sdk.targets -->
  <ItemGroup>
    <_PackageReferenceWithVersion Include="@(PackageReference->HasMetadata('Version'))" />
  </ItemGroup>

  <!-- Pretend this is importing Packages.props -->
  <ItemGroup>
    <PackageReference Update="Foo" Version="2.0.0" />
  </ItemGroup>

  <!-- Add this to Sdk.targets to override any versions set by Packages.props and set them back to their original value from the project -->
  <ItemGroup Condition="'$(AllowPackageReferenceWithVersion)' == 'true'">
    <PackageReference Update="@(_PackageReferenceWithVersion)" Version="%(_PackageReferenceWithVersion.Version)" />
  </ItemGroup>

  <Target Name="Build">
    <Message Text="%(PackageReference.Identity) - %(PackageReference.Version)" Importance="High" />
  </Target>
</Project>
```

yhvicey: Thanks for the example! Patched.

src/CentralPackageVersions/Sdk/Sdk.targets
```diff
@@ -133,6 +139,14 @@
     <PackageReference Include="@(_PackageReferenceWithVersionOverride)"
                       Condition=" '$(EnablePackageVersionOverride)' != 'false' "/>
 
+    <!--
+      Override version of <PackageReference /> items that declared in CentralPackagesFile if AllowPackageReferenceWithVersion
+      is set to true.
+    -->
+    <PackageReference Update="@(_PackageReferenceWithVersion)"
+                      Version="%(_PackageReferenceWithVersion.Version)"
```
jeffkl, line 147: `_PackageReferenceWithVersion` is empty if `AllowPackageReferenceWithVersion` is `true`, so this won't actually do anything right?  Can you update the test to verify that the version is correct which will validate that this new logic is doing what you think it is?

jeffkl: Gentle ping...

yhvicey: Will update in recent days, being busy transferring team & rampup new project

jeffkl:
Closing old PRs, feel free to re-open if you get time to complete this work.

teneko:
Is this still up for a grab, @jeffkl? I use it and find it very useful.

Issue #207, opened 2020-10-07
Microsoft.Build.Traversal does not respect SetPlatform or AdditionalProperties on ProjectReference items


forrestcoward:
The Microsoft.Build.Traversal SDK does not respect `SetPlatform`, `SetConfiguration` or `AdditionalProperties` on `ProjectReference` items, it ignores them, effectively losing the metadata. It is customary for global properties to propagate along `ProjectReference`.

Here is the location I am looking at that is offending: 
https://github.com/microsoft/MSBuildSdks/blob/master/src/Traversal/Sdk/Sdk.targets#L123-L136

```xml
  <Target Name="Build"
          DependsOnTargets="$(BuildDependsOn)"
          Condition=" '$(IsGraphBuild)' != 'true' "
          Returns="@(CollectedBuildOutput)">
    <MSBuild Projects="@(ProjectReference)"
             Condition="'%(ProjectReference.Build)' != 'false'"
             BuildInParallel="$([MSBuild]::ValueOrDefault('%(ProjectReference.BuildInParallel)', '$(BuildInParallel)'))"
             SkipNonexistentProjects="$(SkipNonexistentProjects)"
             SkipNonexistentTargets="$(SkipNonexistentTargets)"
             StopOnFirstFailure="$(StopOnFirstFailure)"
             ContinueOnError="$(ContinueOnError)">
      <Output TaskParameter="TargetOutputs" ItemName="CollectedBuildOutput" />
    </MSBuild>
  </Target>
```

The solution is something like this:
```xml
  <Target Name="Build"
          DependsOnTargets="$(BuildDependsOn)"
          Condition=" '$(IsGraphBuild)' != 'true' "
          Returns="@(CollectedBuildOutput)">
    <MSBuild Projects="@(ProjectReference)"
             Condition="'%(ProjectReference.Build)' != 'false'"
             BuildInParallel="$([MSBuild]::ValueOrDefault('%(ProjectReference.BuildInParallel)', '$(BuildInParallel)'))"
             Properties="%(ProjectReference.SetConfiguration); %(ProjectReference.SetPlatform); %(ProjectReference.AdditionalProperties)"
             SkipNonexistentProjects="$(SkipNonexistentProjects)"
             SkipNonexistentTargets="$(SkipNonexistentTargets)"
             StopOnFirstFailure="$(StopOnFirstFailure)"
             ContinueOnError="$(ContinueOnError)">
[...]

PR #208, opened 2020-10-07
Microsoft.Build.Traversal: propogate well known global properties

forrestcoward:
See issue #207 for discussion on the change.

### DIFF
--- a/src/Traversal/Sdk/Sdk.targets
+++ b/src/Traversal/Sdk/Sdk.targets
@@ -127,6 +127,7 @@
     <MSBuild Projects="@(ProjectReference)"
              Condition="'%(ProjectReference.Build)' != 'false'"
              BuildInParallel="$([MSBuild]::ValueOrDefault('%(ProjectReference.BuildInParallel)', '$(BuildInParallel)'))"
+             Properties="%(ProjectReference.SetConfiguration); %(ProjectReference.SetPlatform); %(ProjectReference.SetTargetFramework)"
              SkipNonexistentProjects="$(SkipNonexistentProjects)"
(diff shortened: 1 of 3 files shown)


src/Traversal/Sdk/Sdk.targets
```diff
@@ -127,6 +127,7 @@
     <MSBuild Projects="@(ProjectReference)"
              Condition="'%(ProjectReference.Build)' != 'false'"
              BuildInParallel="$([MSBuild]::ValueOrDefault('%(ProjectReference.BuildInParallel)', '$(BuildInParallel)'))"
+             Properties="%(ProjectReference.SetConfiguration); %(ProjectReference.SetPlatform); %(ProjectReference.SetTargetFramework)"
```
jeffkl, line 130: Change looks good, do you mind adding a unit test for each property to make sure it doesn't regress?

forrestcoward: Oh, yes!

forrestcoward:
@jeffkl, do I need to update the version in `version.json` as well?

jeffkl:
> do I need to update the version in version.json as well?

Yeah I suppose it changes enough behavior, might as well bump the minor version.  Thanks for suggesting that

src/Traversal/Sdk/Sdk.targets
forrestcoward: I've added tests now. These fail before my change.

jeffkl: I'm looking into it...

jeffkl (approved):
Thanks for the contribution!

PR #216, opened 2020-11-12
Fix issue where Targets are not passed through

### LINKED ISSUE #214: Traversal SDK - Is it possible to invoke a custom target via ProjectReference
Probably more an msbuild question but is it possible to use the Traversal SDK but not invoke the default target but a custom one?

jeffkl:
When a user specifies the Targets metadata on a ProjectReference, they are not passed along to the MSBuild task.  Also added a unit test.

Fixes #214

### DIFF
--- a/src/Traversal.UnitTests/TraversalTests.cs
+++ b/src/Traversal.UnitTests/TraversalTests.cs
@@ -121,6 +121,40 @@ namespace Microsoft.Build.Traversal.UnitTests
         }
 
         [Theory]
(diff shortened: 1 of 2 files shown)


ViktorHofer:
Thanks. Curious, does the Traversal SDK work with Graph Build?

ViktorHofer (approved):

jeffkl:
> Curious, does the Traversal SDK work with Graph Build?

Yes, it should work just fine.

PR #219, opened 2021-01-18
Disable framework references in traversal projects

ViktorHofer:
This should disable restoring targeting packs and adding default references. This comes into play especially on NetCoreApp when setting a custom TFM for traversals.

### DIFF
--- a/src/Traversal/Sdk/Sdk.targets
+++ b/src/Traversal/Sdk/Sdk.targets
@@ -32,6 +32,8 @@
       TargetFramework is required for restore and used to default to .NET Framework v4.5.  However, Traversal projects don't specify a version so it needs to be defaulted here.
     -->
     <TargetFramework Condition="'$(TargetFramework)' == ''">net45</TargetFramework>
+    <!-- We don't want to get any targeting pack references as traversal projects don't build anything. -->
+    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


src/Traversal/Sdk/Sdk.targets
```diff
@@ -32,6 +32,8 @@
       TargetFramework is required for restore and used to default to .NET Framework v4.5.  However, Traversal projects don't specify a version so it needs to be defaulted here.
     -->
     <TargetFramework Condition="'$(TargetFramework)' == ''">net45</TargetFramework>
+    <!-- We don't want to get any targeting pack references as traversal projects don't build anything. -->
+    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
```
jeffkl, line 36: I think this would need to be set in Sdk.props as its read in several places before this .targets file is imported right?

https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props

It should also have a condition just in case a user wants to set it.

```xml
<DisableImplicitFrameworkReferences Condition="'$(DisableImplicitFrameworkReferences)' == ''">true</DisableImplicitFrameworkReferences>
```

ViktorHofer: It's only read by items which are evaluated after properties. But for this SDK it shouldn't matter so I'll move into the props file.

ViktorHofer: done

jeffkl (approved):
Thanks!

PR #220, opened 2021-01-21
Disable framework references in NoTargets projects

jeffkl:
Similar to #219, there's no need to require implicit framework references since NoTargets projects do not produce an assembly.

Also added unit test for Traversal


jeffkl:
FYI @dfederm

PR #225, opened 2021-02-23
Fix extensibility of Compile and CoreCompile

### LINKED ISSUE #223: NoTargets Sdk: CoreCompile is missing from CoreBuildDependsOn
Based on the overrides in the NoTargets SDK, `CoreCompile` is explicitly overridden to "keep extensibility points", but it's missing from `CoreBuildDependsOn`.

jeffkl:
Compile was removed from BuildDependsOn to avoid having it run but this breaks some extensibility points that other SDKs might key off of.  This change brings back the dependency chain to run Compile and CoreCompile but has the bulk of their logic no-op.  This allows the extensibility to continue to work as expected.

Bumped version to 3.0 as this could be a breaking change.

Fixes #223


jeffkl:
FYI @dfederm

PR #230, opened 2021-03-08
Fix #229 AfterTargets used in examples

### LINKED ISSUE #229: NoTargets example should not recommend AfterTargets, can lead to non-determinism in multi-project builds
The example project in [README.md](https://github.com/microsoft/MSBuildSdks/tree/master/src/NoTargets) for the Microsoft.Build.NoTargets Sdk makes use of the AfterTargets property in msbuild:

```xml
 <Target Name="CopyFiles" AfterTargets="Build">
```

This is a bad idea. This target may still be running when when downstream projects finish their ResolveProjectReferences target, leading to race conditions in the build.

Instead, projects should prefer to use the BeforeTargets feature linking to the AfterBuild target. This will ensure their postbuild targets still get run as part of a normal build, that they're run _before_ downstream project begin their BeforeBuild targets, and ensures this target will run after the Compile and CoreBuild targets portion of the build.

aaronla-ms:
Update examples to use `BeforeTargets="AfterBuild"` instead of `AfterTargets="Build"`. `AfterTargets="Build"` can result in race conditions in multi-project repos, as the downstream CoreBuild will be able to proceed before the upstream postbuild steps have completed.

### DIFF
--- a/src/NoTargets/README.md
+++ b/src/NoTargets/README.md
@@ -18,7 +18,7 @@ To have a project that just copies a file:
     <FilesToCopy Include="files\**" />
   </ItemGroup>
 
-  <Target Name="CopyFiles" AfterTargets="Build">
+  <Target Name="CopyFiles" BeforeTargets="AfterBuild">
     <Copy
         SourceFiles="@(FilesToCopy)"
         DestinationFolder="$(OutDir)"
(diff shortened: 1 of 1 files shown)


jeffkl (approved):
Thanks for the contribution!

PR #232, opened 2021-03-20
Update Artifacts assembly path to netcoreapp3.1

circlesabound:
This fixes a FileNotFound exception when using Artifacts with a dotnetcore version < 5.0.

Support for target framework netcoreapp2.1 was dropped in favour of netcoreapp3.1 in #210, but the Robocopy build task here is still referencing a netcoreapp2.1 path which no longer exists. This has been replaced with the equivalent netcoreapp3.1 path.


jeffkl (approved):
Thanks for the contribution!

PR #253, opened 2021-05-25
Fix description to reflect what property the example is for

pergardebrink:
The example description didn't match the example


jeffkl (approved):
Thanks for catching this!

PR #276, opened 2021-08-04
Clear two more items which are read by IDE and NuGet

ViktorHofer:
Clearing these items helps with IDE support but more importantly helps with some niche packaging scenarios when using the NoTargets.Sdk.

@jeffkl this helps with a very specific packaging case that we have in dotnet/runtime when using NoTargets.Sdk but I felt like because the IDE uses these items as well it makes sense to have this in the NoTargets.Sdk.


jeffkl (commented):
Thanks for the contribution!

PR #300, opened 2021-10-09
Clean up logging from NoTargets projects

AArnott:
A NoTargets project will still log (with high importance) something like:

> MyProject ->

Leaving a blank where the primary output would typically go. This looks weird. So this change removes that from being logged. It also disables (because it's the same property that controls it) copying of the primary output that we didn't build.


jeffkl (approved):
Sorry for the delay, GitHub didn't notify me that changes were pushed for some reason.

Thank you for the contribution!

AArnott:
@jeffkl Looks like it failed due to a network outage.

PR #334, opened 2022-01-25
Override GetTargetPath in Traversal SDK

ViktorHofer:
There are use-cases where you want to reference a Traversal SDK project, i.e. when you want to receive its P2P via the `IncludeTransitiveProjectReferences` feature. In such cases msbuild prints out an error when the tree is built with the `--no-dependencies` switch (or I guess when building inside VS) as that will trigger an MSBuild Exec task which calls into the `GetTargetsPath` target which currently either doesn't exist (?) for Traversal projects or returns a TargetPath instead of returning nothing. Hence adding this here to avoid the error that is being printed.

cc @jeffkl

### DIFF
--- a/src/Traversal/Sdk/Sdk.targets
+++ b/src/Traversal/Sdk/Sdk.targets
@@ -235,6 +235,11 @@
              StopOnFirstFailure="$(StopOnFirstFailure)"
              ContinueOnError="$([MSBuild]::ValueOrDefault('$(PublishContinueOnError)', '$(ContinueOnError)'))" />
   </Target>
+  
+  <!--
+    Traversal projects do not build an assembly so dependent projects shouldn't get a path to the target.  Override the GetTargetPath to do nothing.
+  -->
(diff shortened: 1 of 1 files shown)


ViktorHofer:
Hmm unsure why tests are failing

jeffkl:
Does this actually make things work or just get rid of the error about a missing target?  I'm surprised this target doesn't need to return something...

ViktorHofer:
It's not about a missing target but that the not built assembly of the Traversal project is returned when a project references the Traversal project. To answer your question, if the GetTargetPath target doesn't return anything, the non existent build output is not passed down to the caller. This already happens in the NoTargets project: https://github.com/microsoft/MSBuildSdks/blob/27d7541dc309cf7308611f92a6730bc2ddd5f2aa/src/NoTargets/Sdk/Sdk.targets#L74-L82.

jeffkl (approved):
Thanks for the contribution!

PR #338, opened 2022-02-15
Traversal: Have GetTargetPath return collected build outputs

ViktorHofer:
https://github.com/microsoft/MSBuildSdks/commit/1c02e77c5d96c50b803e348b6c3444b578d3dee0 I made the GetTargetPath target for traversal projects return nothing. This is actually wrong as msbuild projects are expected to return the same data in the Build and the GetTargetPath target.
If those two targets aren't in sync, a behavior difference is observable when a traversal project is P2Pd and the dependent project reads from the `CollectedBuildOutput` (which is returned by the `ResolveProjectReferences` target). If the `BuildProjectReferences=true` flag is passed in, the output in the ResolveP2P is currently empty vs. not empty during a normal build without the flag set.

I tested this locally and am also submitting a PR into dotnet/runtime right now which depends on this behavior (currently via the AfterTargets extension point of the Traversal project). Would be great if we could get this into main here so that I can remove the custom implementation in dotnet/runtime.

cc @jeffkl

### DIFF
--- a/src/Traversal/Sdk/Sdk.targets
+++ b/src/Traversal/Sdk/Sdk.targets
@@ -95,6 +95,10 @@
 
     <PublishDependsOn Condition="'$(NoBuild)' == 'true'">
     </PublishDependsOn>
+    
+    <GetTargetPathDependsOn>
+      ResolveP2PReferences;
+    </GetTargetPathDependsOn>
   </PropertyGroup>
 
   <ItemGroup Condition=" '$(TraversalTranslateProjectFileItems)' != 'false' ">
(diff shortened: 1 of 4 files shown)


jeffkl:
@ViktorHofer this is super cool!  Any chance you have time to write some unit tests?  If not, I should have some time today to do it and push to your branch.

ViktorHofer:
To be honest I still haven't familiarized myself with testing msbuild components in neither this repo nor msbuild nor the sdk. At some I definitely want to close that gap but at the moment I would prefer to not spend time on that as I have little time at the moment. If you don't mind Jeff, it would be great if you could add them.

jeffkl:
> To be honest I still haven't familiarized myself with testing msbuild components in neither this repo nor msbuild nor the sdk

No worries, this is definitely the kind of area where the unit testing is a bit of a unicorn since we're exercising MSBuild logic.  That's why I offered, I'll push a commit later today.

jeffkl (approved):

ViktorHofer:
Awesome. Can you please push a new version to nuget? That would be fantastic.

jeffkl:
I am actively working on releasing new bits, the official build is failing with some signing issues so I'm retrying...

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5768594&view=logs&j=141f5c4f-a843-5a00-b45d-c47e8c39a157

src/Traversal/Sdk/Sdk.targets
```diff
@@ -236,10 +238,23 @@
              ContinueOnError="$([MSBuild]::ValueOrDefault('$(PublishContinueOnError)', '$(ContinueOnError)'))" />
   </Target>
   
-  <!--
-    Traversal projects do not build an assembly so dependent projects shouldn't get a path to the target.  Override the GetTargetPath to do nothing.
-  -->
-  <Target Name="GetTargetPath" />
+  <Target Name="GetTargetPath"
```
Youssef1313: @ViktorHofer I think `GetTargetPath` isn't available for a multi targeting build. So if one of the project references is multing-targeting (e.g, `<TargetFrameworks>net9.0;net10</TargetFrameworks>`), this would fail because the project doesn't have the target defined. What should be the right behavior here?

PR #342, opened 2022-03-03
Pass in properties for inner build targets

ViktorHofer:
As the `Clean`, `Test` and `VSTest` targets exist in inner builds, we should pass in properties that `ResolveProjectReferences` in Common.targets passes in as well. The Build and GetTargetPath targets already do so but the above listed targets don't which causes unnecessary clean and test invocations when a specific target framework is set.

cc @jeffkl @ericstj


jeffkl (approved):
Thanks for the contribution!

PR #360, opened 2022-05-12
Set DebugType and DebugSymbols after SDK import

ViktorHofer:
It's fairly common for repositories to set a default DebugType and/or DebugSymbols setting in their Directory.Build.props file. As an example, Arcade does that which impacts every repository in the .NET stack. The NoTargets SDK should overwrite these default settings (including the Microsoft.NET.Sdk setting) as the project clearly doesn't produce a symbol. Setting in the props file to allow a project to overwrite the setting if it explicitly wants to do so.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,7 +6,7 @@ variables:
   BuildConfiguration: 'Debug'
   BuildPlatform: 'Any CPU'
   DotNet3Version: '3.x'
-  DotNet6Version: '6.x'
+  DotNet6Version: '6.0.203'
   MSBuildArgs: '"/p:Platform=$(BuildPlatform)" "/p:Configuration=$(BuildConfiguration)" "/BinaryLogger:$(Build.SourcesDirectory)\$(ArtifactsDirectoryName)\msbuild.binlog"'
   SignType: 'Test'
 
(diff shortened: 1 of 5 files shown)


MeikTranel (approved):

ViktorHofer:
> If everyone's Directory.Build.props conditioned their logic it would be fine but I don't think anyone knows that NoTargets project should be special...

Right. Neither Arcade nor dotnet/runtime nor any other repository that I know of, thinks about this SDK when writing property conditions.

> I'm conflicted on this change. On one hand, this SDK should be a good citizen and not override properties that are already declared.

I understand your concern. This SDK already intentionally overwrites properties, mostly in targets files. As another example, the Traversal SDK overwrites tons of targets that are set by the MSBuild's common logic. I don't think that the NoTargets SDK can continue to exist without overwriting properties / targets set by the Microsoft.NET.Sdk or MSBuild's common files.

I would even go as far as suggest to move these two properties into the SDK's targets file so that props or project level settings are intentionally overwritten.

jeffkl:
Okay I agree, the feature of emitting symbols for a NoTarget is already disabled in other ways

https://github.com/microsoft/MSBuildSdks/blob/0167add5ff5f2d481504ad5d1a8b3b81b8fba402/src/NoTargets/Sdk/Sdk.targets#L44

And there's precedence for setting properties and not letting them be set by the user

https://github.com/microsoft/MSBuildSdks/blob/0167add5ff5f2d481504ad5d1a8b3b81b8fba402/src/NoTargets/Sdk/Sdk.targets#L33

I'm going to update your branch to move the properties to `Sdk.targets` and update the unit tests

jeffkl (approved):
Thanks @ViktorHofer!

PR #386, opened 2022-09-14
Traversal - Remove specific `ProjectReferenceTargets` elements

### LINKED ISSUE #385: Using a custom graph protocol is blocked by the Traversal SDK
When adding a custom protocol, e.g.
```xml
<ProjectReferenceTargets Include="Foo" Targets="Foo" />
```

Only the root traversal project runs the `Foo` target.

I think the problem is this line:
https://github.com/microsoft/MSBuildSdks/blob/96f2169fe2602f2cfdcb86d21a018e35bed8f1b6/src/Traversal/Sdk/Sdk.targets#L40

Perhaps just remove the protocols the Traversal SDK is handling rather than remove all.

aelij:
Fixes #385

### DIFF
--- a/src/Traversal/Sdk/Sdk.targets
+++ b/src/Traversal/Sdk/Sdk.targets
@@ -37,7 +37,13 @@
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


jeffkl (approved):

jeffkl:
I'm working on getting CI running again, give me a few hours...

PR #394, opened 2022-10-10
Error message was incorrect when destination folder was invalid.

MatthieuMEZIL:
There was a bad copy/paste using ItemSpec instead of destination when destination folder was invalid.


jeffkl (approved):

src/Artifacts.UnitTests/ArtifactsTests.cs
```diff
@@ -279,5 +279,39 @@ public void UsingSdkLogic(bool appendTargetFrameworkToOutputPath)
                     }.Select(i => Path.Combine(artifactsPath.FullName, i)),
                     ignoreOrder: true);
         }
+
```
jeffkl, line 282: You'll need to remove this extra line break to get a green build:

```suggestion
```

PR #395, opened 2022-10-15
Add option to disable CopyFilesMarkedCopyLocal target

ViktorHofer:
This adds a new property `SkipCopyFilesMarkedCopyLocal` which can be set to `true` to disable the `CopyFilesMarkedCopyLocal` target.

```xml
<PropertyGroup>
  <SkipCopyFilesMarkedCopyLocal>true</SkipCopyFilesMarkedCopyLocal>
</PropertyGroup>
```

### DIFF
--- a/src/NoTargets/Sdk/Sdk.targets
+++ b/src/NoTargets/Sdk/Sdk.targets
@@ -78,6 +78,9 @@
     The GetTargetPathWithTargetPlatformMoniker target uses a BeforeTargets so the only way to disable it is to override it with an empty target.
   -->
   <Target Name="GetTargetPathWithTargetPlatformMoniker" />
+  
+  <!-- Override CopyFilesMarkedCopyLocal target to not copy references when NoTargetsDoNotReferenceOutputAssemblies is set to false. -->
+  <Target Name="_CopyFilesMarkedCopyLocal" />
 
   <!--
     The GetReferenceAssemblyPaths does not need to run since reference assemblies aren't needed.
--- a/src/NoTargets/Sdk/Sdk.targets
+++ b/src/NoTargets/Sdk/Sdk.targets
@@ -79,8 +79,8 @@
   -->
   <Target Name="GetTargetPathWithTargetPlatformMoniker" />
   
-  <!-- Override CopyFilesMarkedCopyLocal target to not copy references when NoTargetsDoNotReferenceOutputAssemblies is set to false. -->
-  <Target Name="_CopyFilesMarkedCopyLocal" />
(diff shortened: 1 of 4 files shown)


src/NoTargets/Sdk/Sdk.targets
```diff
@@ -78,6 +78,9 @@
     The GetTargetPathWithTargetPlatformMoniker target uses a BeforeTargets so the only way to disable it is to override it with an empty target.
   -->
   <Target Name="GetTargetPathWithTargetPlatformMoniker" />
+  
+  <!-- Override CopyFilesToOutputDirectory target to not copy references when NoTargetsDoNotReferenceOutputAssemblies is set to false. -->
+  <Target Name="CopyFilesToOutputDirectory" />
```
jeffkl, line 83: This will completely disable copying files to the output directory, shouldn't we make this conditional?

ViktorHofer: Correct me if I'm wrong but the NoTargets SDK isn't supposed to copy anything to the intermediate/output directory. We can add an escape hatch but that would require introducing an additional file so that the override is imported.

jeffkl: The primary purpose of NoTargets at the moment is to do everything that a normal CSPROJ/VBPROJ would do except for emitting an assembly.  It currently overrides the CoreCompile target so that the compiler isn't called and sets a bunch of properties to tell the .NET SDK to not try to copy an assembly or symbol file.  The rest of the targets are left in place so that your NoTargets project can place files in the output directory so you can have a project that only stages things.

That said, I'm totally on board with adding a conditional expression to disable all copying to the output directory.  I looked into other ways to disable the target but as I'm sure you found, the only way is to conditionally import a targets file that overrides it.  Let's update this PR to do that, just add a `DisableCopyFilesToOutputDirectory.targets` with this empty target that's imported with a condition like `'$(EnableCopyFilesToOutputDirectory)' == 'false'` (unless we want a more specific property like `NoTargetsDisableCopyFilesToOutputDirectory`).

jeffkl: I've pushed a commit which adds a new property `SkipCopyFilesMarkedCopyLocal` which you can set to `true` to disable the `CopyFilesMarkedCopyLocal` target.  What do you think?

ViktorHofer: Sorry, I didn't see your reply...

ViktorHofer (commented):
Thanks, this looks great!

jeffkl (approved):

PR #419, opened 2023-02-07
Add Microsoft.Build.CopyOnWrite

AndyGerlicher:
Add new Microsoft.Build.CopyOnWrite package to support Copy on Write.


AndyGerlicher:
Most of the code here is copied from MSBuild and minified. The Copy task is fairly complicated, and I wanted to maintain as much compatibility as possible(error codes, etc.).

src/CopyOnWrite/Copy.cs
```diff
@@ -0,0 +1,1044 @@
...
+        public static readonly int CopyTaskParallelism = ParseIntFromEnvironmentVariableOrDefault("MSBUILDCOPYTASKPARALLELISM", -1);
+
+        private static int ParseIntFromEnvironmentVariableOrDefault(string environmentVariable, int defaultValue)
+        {
+            return int.TryParse(Environment.GetEnvironmentVariable(environmentVariable), out int result)
+                ? result
+                : defaultValue;
+        }
+
+        #region CopyOnWrite functionality
```
AndyGerlicher, line 978: This is the majority of my change to the Copy task (till the end of file).

jeffkl: Hmm in a perfect world, we would update the stock `Copy` task to allow us to override it.  But then none of this would work with older versions of MSBuild 😢 Are you going to keep this up-to-date when the `Copy` task changes? 😄

src/CopyOnWrite/Copy.cs
```diff
@@ -0,0 +1,1044 @@
...
+
+        private static readonly Dictionary<string, bool> _reFsDrives = new(StringComparer.OrdinalIgnoreCase);
+
+        /// <summary>
+        /// Check for CopyOnWrite support. Result is cached by drive root.
+        /// </summary>
+        /// <param name="source">Full path to source file.</param>
+        /// <param name="dest">Full path to destination file.</param>
+        /// <returns>True when CopyOnWrite appears to be supported.</returns>
+        private bool GuessCopyOnWriteSupported(string source, string dest)
```
jeffkl, line 1011: `Guess`?  What is this, ChatGPT?

src/CopyOnWrite/build/Microsoft.Build.CopyOnWrite.targets
```diff
@@ -0,0 +1,6 @@
+﻿<Project>
+    <UsingTask
```
jeffkl, line 2: Have you tested this?  I helped Erik a while back and we discovered that for `<UsingTask />` declaration, the first one wins.  Maybe this is still first because the `Copy` task comes from the `.tasks` file in MSBuild?  I'm not sure the import order for that one.

I think it would still be safer for this to be in the `.props`.

AndyGerlicher: I tested this as both an SDK or PackageReference.

https://github.com/dotnet/msbuild/blob/446f42e6d1478acb44faec787f90f57f38c08b73/src/Build/BackEnd/TaskExecutionHost/TaskExecutionHost.cs#L811

Based on that, declared anywhere in the Project wins over the *.tasks file in the MSBuild directory. This feels like more of a `.targets` thing, and I already tested that way. But I don't feel strongly about it.

erikma (approved):

PR #423, opened 2023-02-23
Fix NoTargets example for build incrementalism

### LINKED ISSUE #422: [NoTargets] FileWrites has no impact on build incrementalism
Taking the repo's [example](https://github.com/microsoft/MSBuildSdks/blob/main/src/NoTargets/README.md#example) for repro, the items written in `FileWrites` are created too late for `_CleanGetCurrentAndPriorFileWrites` to be processed. Is this an issue with the example (fine-tune Before/AfterTargets) or with the NoTargets SDK?

Here's the graphical representation of the order of actions:

image

stan-sz:
Fixes #422

### DIFF
--- a/src/NoTargets/README.md
+++ b/src/NoTargets/README.md
@@ -18,7 +18,7 @@ To have a project that just copies a file:
     <FilesToCopy Include="files\**" />
   </ItemGroup>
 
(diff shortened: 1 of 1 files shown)


jeffkl (approved):
Thanks!

PR #429, opened 2023-03-13
Fix project reference protocol for outer build nodes

DmitriyShepelev:
Semi-related issue: https://github.com/dotnet/msbuild/issues/4344

Fix project reference protocol for outer build nodes. This change ensures that the set of targets to call for the outer build of a multitargeting node referenced by a `Microsoft.Build.Traversal` SDK-importing `dirs.proj` contains the default targets.

Before:
image

After:
f0ae77a7-1c89-4fac-a847-53b73579d236


dfederm (approved):

src/Traversal/Sdk/Sdk.targets
```diff
@@ -45,7 +45,9 @@
     <ProjectReferenceTargets Remove="Publish" />
     <ProjectReferenceTargets Remove="Pack" />
 
+    <ProjectReferenceTargets Include="Build" Targets=".default" OuterBuild="true" />
     <ProjectReferenceTargets Include="Build" Targets=".default" />
```
dfederm, line 49: I actually wonder, should this be removed then? Or would that break non-multitargeting projects?

jeffkl: I'm assuming the `OuterBuild="true"` is good enough?  Have you looked at how static graph API handles this attribute?

DmitriyShepelev: `<ProjectReferenceTargets Include="Build" Targets=".default" />` would still be necessary to ensure the default targets are called on referenced non-multitargeting projects.
https://github.com/dotnet/msbuild/blob/main/documentation/specs/static-graph.md#multitargeting-details

In the case I'm addressing, the item wouldn't be necessary though.

jeffkl (approved):

PR #430, opened 2023-03-23
Avoid restoring net45 targeting pack in Traversal SDK

ViktorHofer:
The NoTargets SDK already sets the AutomaticallyUseReferenceAssemblyPackages property to false but the Traversal SDK didn't. As the Traversal SDK defines a default TFM (net45), the reference assembly pack is being downloaded automatically. This shows us as a prebuild in dotnet/runtime and other repositories that build from source.

cc @NikolaMilosavljevic

### DIFF
--- a/src/Traversal/Sdk/Sdk.props
+++ b/src/Traversal/Sdk/Sdk.props
@@ -8,6 +8,9 @@
 
   <PropertyGroup>
     <UsingMicrosoftTraversalSdk>true</UsingMicrosoftTraversalSdk>
+    
+    <!-- Don't automatically reference assembly packages since NoTargets don't need reference assemblies -->
+    <AutomaticallyUseReferenceAssemblyPackages Condition="'$(AutomaticallyUseReferenceAssemblyPackages)' == ''">false</AutomaticallyUseReferenceAssemblyPackages>
   </PropertyGroup>
 
   <Import Project="$(CustomBeforeTraversalProps)" Condition=" '$(CustomBeforeTraversalProps)' != '' And Exists('$(CustomBeforeTraversalProps)') " />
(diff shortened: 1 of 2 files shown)


ViktorHofer:
@jeffkl @AndyGerlicher any idea why CI is failing? Unsure if that's related to my change as it complains about a file not found. I wouldn't expect the current Traversal tests to depend on the .NET Framework reference assemblies.

jeffkl:
Yeah sorry that's a weird error that MSBuild throws occasionally, I haven't had time to root cause it.

src/Traversal/Sdk/Sdk.props
```diff
@@ -31,6 +31,9 @@
     
     <!-- Targeting packs shouldn't be referenced as traversal projects don't compile. -->
     <DisableImplicitFrameworkReferences Condition="'$(DisableImplicitFrameworkReferences)' == ''">true</DisableImplicitFrameworkReferences>
+    
+    <!-- Don't automatically reference assembly packages since NoTargets don't need reference assemblies -->
+    <AutomaticallyUseReferenceAssemblyPackages Condition="'$(AutomaticallyUseReferenceAssemblyPackages)' == ''">false</AutomaticallyUseReferenceAssemblyPackages>
```
jeffkl, line 36: I'm going to add a unit test real quick...

jeffkl: Wow yeah good thing I did that, this is setting it after the .NET SDK so it has no affect since its set to `true` by the SDK.  I'm going to move this up to before the .NET SDK is imported.

jeffkl (approved):

jeffkl:
> After merging the change in, can we please publish a new version to nuget.org?

Yes absolutely, I usually publish a package for every change

Issue #431, opened 2023-03-24
Are the msbuild sdk nuget package dependencies necessary


ViktorHofer:
While looking at the latest available Traversal package on nuget.org, I noticed that the package contains empty dependency groups for `net40` and `netstandard2.0`. While this is just for cosmetics, I wonder if we want to omit them instead by setting `<SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>` in the project files.

Happy to submit a PR.

jeffkl:
Technically it doesn't matter since `Microsoft.Build.Traversal` is a NuGet-based MSBuild Project SDK and during "restore" of this kind of package, NuGet does not respect dependencies or target framework compatibility.  I'm pretty sure you can't add a `<PackageReference />` to that kind of package either.

But if you want to submit a PR, I'll gladly accept it 😄

ViktorHofer:
> Technically it doesn't matter since Microsoft.Build.Traversal is a NuGet-based MSBuild Project SDK and during "restore" of this kind of package, NuGet does not respect dependencies or target framework compatibility.

While NuGet doesn't respect it, we do have some tools that inspect packages and their dependencies, i.e. ApiCompat, or [SBRP](https://github.com/dotnet/source-build-reference-packages). I will submit a PR to suppress the dependencies.

PR #436, opened 2023-04-13
Suppress package dependencies in msbuild sdk packages

ViktorHofer:
Fixes https://github.com/microsoft/MSBuildSdks/issues/431

Unrelated question: Why do the Traversal and the NoTargets projects target net40?


jeffkl (approved):
Thanks for the contribution!

ViktorHofer:
> In order to load the projects in Visual Studio, you need to specify a target framework. I think at the time I just picked the lowest one? The target framework shouldn't be used anywhere so it could probably be anything.

Sorry, I should have elaborated. I meant the build target framework in this repository for the projects: `<TargetFrameworks>net40;netstandard2.0</TargetFrameworks>`.

PR #437, opened 2023-04-15
Add Artifact file renaming feature

MatthieuMEZIL:
Add Artifact file renaming feature

### DIFF
--- a/src/Artifacts.UnitTests/ArtifactsTests.cs
+++ b/src/Artifacts.UnitTests/ArtifactsTests.cs
@@ -356,5 +356,81 @@ namespace Microsoft.Build.Artifacts.UnitTests
                     }.Select(i => Path.Combine(artifactsPath.FullName, i)),
                     ignoreOrder: true);
         }
+
+        [Fact]
+        public void RenamedFiles()
+        {
+            CreateFiles(
+                Path.Combine("bin", "Debug"),
+                "foo.exe",
+                "foo.pdb",
+                "foo.exe.config",
+                "bar.dll",
+                "bar.pdb",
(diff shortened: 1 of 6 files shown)


jeffkl:
@MatthieuMEZIL can you add some examples in the pull request description on how this will work for end users?  I'm having trouble picturing how to convey the intent to have files renamed.

MatthieuMEZIL:
@jeffkl e.g.
<Artifact
            Include="App.config"
            RenamedFiles="Foo.exe.config"
            DestinationFolder="$(O)" />

jeffkl:
@MatthieuMEZIL a lot of people use wildcards when staging artifacts:

```xml
<ItemGroup>
  <Artifact Include="MyFolder\MyStuff"
            FileMatch="*.txt *.ini *.xml"
            DestinationFolder="$(ArtifactsPath)" />
</ItemGroup>
```

If there's more than file found in the bucket, should there be an error?  We don't want to copy all files that match to the same destination file.

I think ideally users could supply a transform but that would be very complicated because we'd have to let the user supply a pattern with tokens and get it right:
```xml
<ItemGroup>
  <Artifact Include="MyFolder\MyStuff"
            FileMatch="*.txt *.ini *.xml"
            DestinationFolder="$(ArtifactsPath)"
            DestinationFile="$filename$.altered.$extension$" />
</ItemGroup>
```

Even then, I think there should be an error or warning if the same file was written to the same path more than once.  I'm wondering if its better for the build logic to handle this rename and then have Artifacts just stage the outputs from there after the files are already renamed?

dfederm:
Closing stale PR

PR #439, opened 2023-04-26
Add ability to skip projects dynamically in Traversal projects and Visual Studio solution files

jeffkl:
This implements the ability to dynamically skip projects when building a Traversal project or Visual Studio solution file.  Users need to implement a target in `Directory.Build.targets`, the project itself, or some other import:

This example skips VSIX projects when building with `dotnet build`:
```xml
<Target Name="ShouldSkipProject" Returns="@(ProjectToSkip)">
  <ItemGroup Condition="'$(MSBuildRuntimeType)' == 'Core'">
    <!-- Skip building Visual Studio Extension (VSIX) projects if the user is building with dotnet build since its only supported to build those projects with MSBuild.exe -->
    <ProjectToSkip Include="$(MSBuildProjectFullPath)"
                   Message="Visual Studio Extension (VSIX) projects cannot be built with dotnet.exe and require you to use msbuild.exe or Visual Studio."
                   Condition="'$(VsSDKVersion)' != ''" />
  </ItemGroup>
</Target>
```

Users can also now extend Visual Studio solution file builds with Microsoft.Build.Traversal to get the same functionality.  To do this, they must create a `Directory.Solution.props` and `Directory.Solution.targets` next to the solution file:

Directory.Solution.props:
```xml
<Project>
  <Import Project="Microsoft.Build.Traversal" Project="Sdk.props" />
</Project>
```

Directory.Solution.targets:
```xml
<Project>
  <Import Project="Microsoft.Build.Traversal" Project="Sdk.targets" />
</Project>
```

This gives them the ability to dynamically skip projects in the same way.


avivanoff:
@jeffkl, this looks like a breaking change when upgrading to 4.0.0 as it requires defining a ShouldSkipProject target even if you do not want to skip anything. Why not define an empty ShouldSkipProject target in the SDK?

PR #458, opened 2023-06-21
Disable Artifacts if project is using .NET SDK's built-in artifacts functionality

jeffkl:
If a user has opted into the .NET SDK's artifacts functionality by setting the MSBuild property `UseArtifactsOutput`, then disable Microsoft.Build.Artifacts since you would only want to be using one or the other.

Also enables the pipeline to run tests against .NET 7.0 and .NET 8 while no longer running against .NET Core 3

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -8,5 +8,7 @@
     <NoWarn>$(NoWarn);NU5128;SA0001</NoWarn>
     <RestoreUseStaticGraphEvaluation>true</RestoreUseStaticGraphEvaluation>
     <UseArtifactsOutput>false</UseArtifactsOutput>
+    <IsTestProject Condition="$(MSBuildProjectName.EndsWith('UnitTests'))">true</IsTestProject>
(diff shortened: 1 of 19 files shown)


jeffkl:
Related PR in .NET SDK to disable itself if Microsoft.Build.Artifacts is in use: https://github.com/dotnet/sdk/pull/33470

dsplaisted (approved):

src/Shared/MSBuildSdkTestBase.cs
```diff
@@ -47,7 +47,7 @@ public void Dispose()
 
         protected DirectoryInfo CreateFiles(string directoryName, params string[] files)
         {
-            DirectoryInfo directory = new DirectoryInfo(Path.Combine(TestRootPath, directoryName));
+            DirectoryInfo directory = Directory.CreateDirectory(Path.Combine(TestRootPath, directoryName));
```
erikmav, line 50: >Directory.CreateDirectory [](http://example.com/codeflow?start=38&length=25)

Should not create the dir here, it's important for Artifacts testing to ensure that Artifacts creates the destination dir.

Issue #453, opened 2023-06-09
CopyOnWrite: winerror -2144272384 when drive is using biltocker


PiDiBi:
getting `Failed retrieving volume information for E:\ with winerror -2144272384` when used on drive with bitlocker
not sure if you would be able to reproduce it but some try catch would be nice, so when detected this error to fall back to "no ReFS" case

**global.json**
`"Microsoft.Build.CopyOnWrite" : "1.0.240"`

**Directory.Build.Targets**
```xml
<Project>
  <Sdk Name="Microsoft.Build.CentralPackageVersions" />
  <Sdk Name="Microsoft.Build.CopyOnWrite" />
```

**Error**
```bash
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(5215,5): error MSB4061: The "Copy" task could not be instantiated from "C:\NUGET_PACKAGES\microsoft.build.copyonwrite\1.0.240\Sdk\..\lib\netstandard2.0\Microsoft.Build.CopyOnWrite.dll".  [C:\wagit2\AAPT-Antares-WebSites\src\Hosting\System32Tools\FirstPartyControl\FirstPartyControl.csproj]
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(5215,5): error MSB4061: System.TypeInitializationException: The type initializer for 'Microsoft.Build.Tasks.Copy' threw an exception. ---> System.ComponentModel.Win32Exception: Failed retrieving volume information for E:\ with winerror -2144272384 [C:\wagit2\AAPT-Antares-WebSites\src\Hosting\System32Tools\FirstPartyControl\FirstPartyControl.csproj]
```

PiDiBi:
error is coming from here https://github.com/microsoft/CopyOnWrite/blob/e1e9f45557009e04e85a1c011245258cb73e6e9f/lib/Windows/VolumeInfoCache.cs#L135
but IMHO should be sunk and logged here
https://github.com/microsoft/MSBuildSdks/blob/14bd37933288ab06fd303fc3752616bec4813ef4/src/CopyOnWrite/Copy.cs#LL1026C13-L1036C27 and false returned in case of exception

PiDiBi:
created PR #454

PR #462, opened 2023-07-11
Add parallelism and copy-on-write links to Robocopy task

erikmav:
- Use the CopyOnWrite library and an action block to greatly speed up artifact copies. Use parallelism settings similar to those in the Microsoft.Build.CopyOnWrite SDK and allow turning off CoW linking with a task setting.
- Fix an overcopying bug where the same source:destination pair would be copied multiple times.
- Don't copy files onto themselves (seen in occasional cases).
- Linearize same-destination copies from different sources.
- Update major version of Artifacts.
- Unify on 'DisableCopyOnWrite' property for Copy and Artifacts
- Migrate CoW SDK Copy task to use CoW package for clone file compat checks.
- Fix possibility of under-copying different-cased files/paths on Linux by varying path comparers by OS. NOTE: Did not update the CoW Copy task, and MSBuild itself seems to use IgnoreCase in most cases. Need feedback here.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,6 +4,7 @@
     <MicrosoftBuildPackageVersion>17.6.3</MicrosoftBuildPackageVersion>
   </PropertyGroup>
   <ItemGroup>
+    <PackageVersion Include="CopyOnWrite" Version="0.3.6" Condition=" '$(TargetFramework)' != 'net46' " />
     <PackageVersion Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildPackageVersion)" />
     <PackageVersion Update="Microsoft.Build.Utilities.Core" Version="16.9.0" Condition="'$(TargetFramework)' == 'netcoreapp3.1' Or '$(TargetFramework)' == 'netstandard2.0'" />
     <PackageVersion Update="Microsoft.Build.Utilities.Core" Version="15.9.20" Condition="'$(TargetFramework)' == 'net46'" />
@@ -15,9 +16,10 @@
     <PackageVersion Include="MSBuild.ProjectCreation" Version="10.0.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="Shouldly" Version="4.2.1" />
+    <PackageVersion Include="System.Threading.Tasks.Dataflow" Version="6.0.0" Condition=" '$(TargetFramework)' != 'net46' "/>
+    <PackageVersion Include="System.Threading.Tasks.Dataflow" Version="4.11.1" Condition=" '$(TargetFramework)' == 'net46' "/>
     <PackageVersion Include="xunit" Version="2.4.2" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.4.5" />
-    <PackageVersion Include="CopyOnWrite" Version="0.3.2" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Artifacts.UnitTests/RobocopyTests.cs
+++ b/src/Artifacts.UnitTests/RobocopyTests.cs
@@ -7,14 +7,32 @@ using Microsoft.Build.Framework;
 using Microsoft.Build.UnitTests.Common;
 using Microsoft.Build.Utilities.ProjectCreation;
 using Shouldly;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using Xunit;
+
+#nullable enable
 
 namespace Microsoft.Build.Artifacts.UnitTests
 {
     public class RobocopyTests : MSBuildSdkTestBase
     {
+        private static readonly bool IsWindows = Environment.OSVersion.Platform == PlatformID.Win32NT;
+
+        [Fact]
+        public void DedupKeyOsDifferences()
+        {
+            var lowercase = new Robocopy.CopyFileDedupKey("foo", "bar");
+            var uppercase = new Robocopy.CopyFileDedupKey("FOO", "BAR");
+            Robocopy.CopyFileDedupKey.ComparerInstance.Equals(lowercase, uppercase).ShouldBe(IsWindows);
+            (Robocopy.CopyFileDedupKey.ComparerInstance.GetHashCode(lowercase) ==
+             Robocopy.CopyFileDedupKey.ComparerInstance.GetHashCode(uppercase)).ShouldBe(IsWindows);
+        }
+
         [Fact]
         public void NonRecursiveWildcards()
         {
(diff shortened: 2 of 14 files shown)


src/Artifacts.UnitTests/RobocopyTests.cs
```diff
@@ -181,11 +184,11 @@ public void SingleFileMatchRecursive()
                 Path.Combine("baz", "baz.dll"),
                 Path.Combine("foo", "foo", "foo", "foo.pdb"));
 
-            DirectoryInfo destination = new DirectoryInfo(Path.Combine(TestRootPath, "destination"));
+            var destination = new DirectoryInfo(Path.Combine(TestRootPath, "destination"));
```
jeffkl, line 187: I hate `var` but its not banned.  Be aware that I'll probably undo this in some future commit...

erikmav: Restored in next push.

erikmav:
@jeffkl I don't see how to increment the Microsoft.Build.CopyOnWrite SDK version - ideally the newer 0.3.3 CopyOnWrite should be used in a new version of that package.

jeffkl:
> @jeffkl I don't see how to increment the Microsoft.Build.CopyOnWrite SDK version - ideally the newer 0.3.3 CopyOnWrite should be used in a new version of that package.

Would just setting this to 1.1 be enough? https://github.com/microsoft/MSBuildSdks/blob/main/src/CopyOnWrite/version.json

The build number automatically increments for each commit but I manually adjust the minor version for feature releases.

erikmav:
I think 1.0.<new build number> would be fine as long as it gets published after this commit. Not sure how that works for this repo.

erikmav:
One other note - tests ran successfully on a Dev Drive with CoW capability.

Directory.Packages.props
```diff
@@ -2,8 +2,12 @@
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <MicrosoftBuildPackageVersion>17.6.3</MicrosoftBuildPackageVersion>
+
+    <DataflowPackageVersion Condition="'$(TargetFramework)' != 'net46'">6.0.0</DataflowPackageVersion>
+    <DataflowPackageVersion Condition="'$(TargetFramework)' == 'net46'">4.11.1</DataflowPackageVersion>
```
AndyGerlicher, line 7: This won't work for all projects so it could cause issues. The condition needs to be on the Item below, not on the property.

src/Artifacts/FileSystem.cs
```diff
@@ -59,5 +66,27 @@ public bool FileExists(FileInfo file)
         {
             return file.Exists;
         }
+
+        public bool TryCloneFile(FileInfo sourceFile, FileInfo destinationFile)
+        {
+#if NETSTANDARD2_0_OR_GREATER
+            string sourcePath = sourceFile.FullName;
+            string destPath = destinationFile.FullName;
+            if (CoW.CopyOnWriteLinkSupportedBetweenPaths(sourcePath, destPath))
```
AndyGerlicher, line 75: Most users this will be false 100% of the time. Is there a perf hit? I implemented a different approach to avoid checking each time. If this is truly no perf hit we should change the other SDK to match this approach (which is better). But given the perf requirements, I would guess this would be a perf hit on large repos.

erikmav: The CoW package has a highly tuned, cached lookup table to minimize time to make this call (as I expected it to be used on most calls as we're doing here). It's more complete (handles more cases) than the drive letter cache in the CoW SDK in this repo. I can migrate that logic to use this.

src/CopyOnWrite/Copy.cs
```diff
@@ -1010,38 +1009,17 @@ private bool TryCopyOnWrite(string source, string dest)
...
-                    {
-                        var supportsCloneFile = _cow.CopyOnWriteLinkSupportedInDirectoryTree(key);
-                        Log.LogMessage(MessageImportance.Low,
-                            supportsCloneFile
-                                ? $"Drive {key} has support for CloneFile. All copies will attempt to use CloneFile."
-                                : $"Drive {key} does not have support for CloneFile. File.Copy will be used.");
-                        return supportsCloneFile;
-                    })).Value;
-            }
-            catch (Exception ex)
```
AndyGerlicher, line 1040: The try/catch was added to fix #453. Should probably just leave it?

src/CopyOnWrite/Copy.cs
```diff
@@ -976,7 +976,8 @@ private static int ParseIntFromEnvironmentVariableOrDefault(string environmentVa
         }
 
         #region CopyOnWrite functionality
-        private static readonly ICopyOnWriteFilesystem _cow = CopyOnWriteFilesystemFactory.GetInstance();
+        private static readonly ICopyOnWriteFilesystem CoW = CopyOnWriteFilesystemFactory.GetInstance();
+        private static readonly bool DisableCoW = Environment.GetEnvironmentVariable("MSBUILDDISABLECOPYONWRITE") is not null;
```
AndyGerlicher, line 980: This is already toggled at the task level:
https://github.com/microsoft/MSBuildSdks/blob/c8f848540db223df28b3343588111592c9eb2e8a/src/CopyOnWrite/build/Microsoft.Build.CopyOnWrite.targets#L5C14-L5C14

I probably wouldn't do both. If you do want this, you should also probably check the value so `MSBUILDDISABLECOPYONWRITE=0` doesn't disable it, for example.

src/CopyOnWrite/Copy.cs
erikmav: Will restore. Interesting case in the bug, will add that to the CoW library as well

src/CopyOnWrite/Copy.cs
erikmav: WIll remove here. There's another copy of this code in Artifacts, I'll either remove and add a flag to the targets file, or keep and improve the comparison

AndyGerlicher: I think it's the right approach for artifacts though. For Copy, if you opt out it should fallback to the old task entirely not use the "new" task with the feature disabled.

Although there should probably be a single way to opt-out for both, But to do that you would have to pass in the property to the artifacts Task.

erikmav: Added param to the Robocopy task code and updated RobocopyFiles and CopyArtifacts targets to use EnableCopyOnWriteWin

src/CopyOnWrite/Copy.cs
erikmav: Added that case to the CoW package in 0.3.4. (0.3.5 in latest commit sets the AssemblyVersion to 0.0.0.1 to avoid collisions if a repo includes differing versions of Microsoft.Build.CopyOnWrite and Microsoft.Build.Artifacts.)

erikmav:
Pending: Stress testing showing occasional breaks vs. single-threaded implementation when there are multiple `Robocopy` items copying different files to the same destination. Need parallel partitioning by destination file to allow this case to avoid having to have repos fix up complex recursive copies.

Issue #463, opened 2023-07-31
Microsoft.Build.Artifacts 5.0.0 broken on Linux


timmydo:
Error changing from 4.2.0 to 5.0.0:

```
/home/timmy/.nuget/packages/microsoft.build.artifacts/5.0.0/build/Microsoft.Build.Artifacts.Common.targets(51,5): error MSB4061: The "Robocopy" task could not be instantiated from "/home/timmy/.nuget/packages/microsoft.build.artifacts/5.0.0/build/netstandard2.0/Microsoft.Build.Artifacts.dll".  [blah.csproj]
/home/timmy/.nuget/packages/microsoft.build.artifacts/5.0.0/build/Microsoft.Build.Artifacts.Common.targets(51,5): error MSB4061: System.TypeInitializationException: The type initializer for 'Microsoft.Build.Artifacts.FileSystem' threw an exception. [blah.csproj]
/home/timmy/.nuget/packages/microsoft.build.artifacts/5.0.0/build/Microsoft.Build.Artifacts.Common.targets(51,5): error MSB4061:  ---> System.IO.FileNotFoundException: Could not load file or assembly 'CopyOnWrite, Version=0.9.9999.0, Culture=neutral, PublicKeyToken=49642b47d27a30e9'. The system cannot find the file specified. [blah.csproj]
/home/timmy/.nuget/packages/microsoft.build.artifacts/5.0.0/build/Microsoft.Build.Artifacts.Common.targets(51,5): error MSB4061:  [blah.csproj]
/home/timmy/.nuget/packages/microsoft.build.artifacts/5.0.0/build/Microsoft.Build.Artifacts.Common.targets(51,5): error MSB4061: File name: 'CopyOnWrite, Version=0.9.9999.0, Culture=neutral, PublicKeyToken=49642b47d27a30e9' [blah.csproj]
/home/timmy/.nuget/packages/microsoft.build.artifacts/5.0.0/build/Microsoft.Build.Artifacts.Common.targets(51,5): error MSB4061:    --- End of inner exception stack trace --- [blah.csproj]
/home/timmy/.nuget/packages/microsoft.build.artifacts/5.0.0/build/Microsoft.Build.Artifacts.Common.targets(51,5): error MSB4061:    at Microsoft.Build.Artifacts.FileSystem.get_PathComparer() [blah.csproj]
[...]

PR #468, opened 2023-08-01
Fix error on missing CoW assembly in Artifacts

erikmav:
Declare the CoW package private and explicitly copy its assembly alongside the Artifacts assembly for net472 and netstandard2.0. For #463

### DIFF
--- a/src/Artifacts/version.json
+++ b/src/Artifacts/version.json
@@ -1,4 +1,4 @@
 ﻿{
   "inherit": true,
-  "version": "5.0"
+  "version": "5.0.1"
 }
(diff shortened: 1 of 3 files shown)


jeffkl (approved):
:shipit:

src/Artifacts/version.json
```diff
@@ -1,4 +1,4 @@
 ﻿{
   "inherit": true,
-  "version": "5.0"
```
jeffkl, line 3: Revert this, the version is calculated automatically and I only increment the minor version for new features and major version for breaking changes.  Build number is just git height and used for fixes.

jeffkl (approved):

PR #473, opened 2023-08-28
add runtests

novacole:


src/RunTests/build/Microsoft.Build.RunTests.targets
```diff
@@ -0,0 +1,12 @@
+<Project>
+  <UsingTask TaskName="Microsoft.Build.RunTestsTask" AssemblyFile="$(MSBuildThisFileDirectory)netstandard2.0\Microsoft.Build.RunTests.dll" />
+  <Target Name="RunTests" AfterTargets="MSBuildRunTests" Condition="$(IsTestProject)==true">
```
MarcoRossignoli, line 3: is we want to plug a different test runner we can do it using a different `Condition` here right?

src/RunTests/RunTestsTask.cs
```diff
@@ -0,0 +1,443 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Build
+{
+    public class RunTestsTask : Build.Utilities.Task
```
MarcoRossignoli, line 13: I would rename in `RunVSTestTask`

src/RunTests/Microsoft.Build.RunTests.csproj
```diff
@@ -0,0 +1,31 @@
...
+		<NoDefaultExcludes>true</NoDefaultExcludes>
+		<BuildOutputTargetFolder>build\</BuildOutputTargetFolder>
+		<AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
+		<DevelopmentDependency>true</DevelopmentDependency>
+		<LangVersion>latest</LangVersion>
+		<SymbolPackageFormat>snupkg</SymbolPackageFormat>
+	</PropertyGroup>
+	<ItemGroup>
+		<PackageReference Include="Microsoft.Build.Utilities.Core" />
+		<PackageReference Include="Microsoft.TestPlatform" />
```
MarcoRossignoli, line 16: I wonder if we need it, means take always this deps also if in future we add new runners

src/RunTests/README.md
```diff
@@ -0,0 +1,30 @@
...
+  </ItemGroup>
+</Project>
+```
+This example will include the `Microsoft.Build.RunTests` task for all NuGet-based projects in your repo.
+
+## Example
+To run tests
+```
+$env:MSBUILDENSURESTDOUTFORTASKPROCESSES=1
+msbuild /nodereuse:false /t:Test
```
MarcoRossignoli, line 23: How does it work if I have a `Test` target defined plus `MSBuildRunTests`?

MarcoRossignoli:
cc: @nohwnd for review

src/RunTests/build/Microsoft.Build.RunTests.targets
novacole: Yes, that is correct!

src/RunTests/README.md
novacole: Updated the readme to reflect the target name 'MsBuildRunTests', instead of 'Test'.

src/RunTests/build/Microsoft.Build.RunTests.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+	<PropertyGroup>
+		<NugetPath Condition="$(NUGET_PACKAGES)==''">$(USERPROFILE)\.nuget</NugetPath>
+		<NugetPath Condition="$(NUGET_PACKAGES)!=''">$(NUGET_PACKAGES)</NugetPath>
+		<VSTestRunnerVersion Condition="$(VSTestRunnerVersion)==''">17.6.3</VSTestRunnerVersion>
```
dfederm, line 5: Why this specific version? Does this mean that people will need to coordinate updating their VS instance with updating this package?

src/RunTests/build/Microsoft.Build.RunTests.targets
```diff
@@ -0,0 +1,12 @@
...
+    <RunTestsTask VSTestRunnerVersion="$(VSTestRunnerVersion)" TestFileFullPath="$(TargetPath)" VSTestSetting="$(VSTestSetting)" VSTestTestAdapterPath="$(VSTestTestAdapterPath)"
+      VSTestFramework="$(VSTestFramework)" VSTestPlatform="$(VSTestPlatform)" VSTestTestCaseFilter="$(VSTestTestCaseFilter)" VSTestLogger="$(VSTestLogger)"
+      VSTestResultsDirectory="$(VSTestResultsDirectory)" VSTestVerbosity="$(VSTestVerbosity)" VSTestCollect="$(VSTestCollect)" VSTestBlame="$(VSTestBlame)"
+      VSTestBlameCrash="$(VSTestBlameCrash)" VSTestBlameCrashDumpType="$(VSTestBlameCrashDumpType)" VSTestBlameCrashCollectAlways="$(VSTestBlameCrashCollectAlways)"
+      VSTestBlameHang="$(VSTestBlameHang)" VSTestBlameHangDumpType="$(VSTestBlameHangDumpType)" VSTestTraceDataCollectorDirectoryPath="$(VSTestTraceDataCollectorDirectoryPath)"
+      VSTestNoLogo="$(VSTestNoLogo)" VSTestArtifactsProcessingMode="$(VSTestArtifactsProcessingMode)" VSTestSessionCorrelationId="$(VSTestSessionCorrelationId)"
+      SkipExecution="$(CacheHit)" NugetPath="$(NugetPath)" />
```
dfederm, line 10: > SkipExecution="$(CacheHit)"

Not sure what "CacheHit" is here. This SDK doesn't seem to do anything with caching.

src/RunTests/README.md
```diff
@@ -0,0 +1,30 @@
...
+To run tests
+```
+$env:MSBUILDENSURESTDOUTFORTASKPROCESSES=1
+msbuild /nodereuse:false /t:MSBuildRunTests
+```
+
+To build and run tests
+```
+$env:MSBUILDENSURESTDOUTFORTASKPROCESSES=1
+msbuild /nodereuse:false /t:Build;MSBuildRunTests
```
nohwnd, line 29: Integration tests (at least ours) depend on nuget packages being built, and not all our projects generate packages during Build.

How is the order going to work when I also want to call Pack? Will the new task run before or after, or there is no guarantee? Ideally it would run after, but without taking dependency on Pack. Can this be achieved?

src/RunTests/RunTestsTask.cs
nohwnd: I agree with this suggestion, the whole task and target should use VSTest in the name.

src/RunTests/build/Microsoft.Build.RunTests.targets
nohwnd, line 3: Would it make sense to name this VSTest after the existing task we have in dotnet sdk? I don't know if this task will replace the on in the dotnet sdk, but if user has existing pipeline in place that is using `/t:VSTest`, we should not break them.

`<Target Name="VSTest" DependsOnTargets="ShowInfoMessageIfProjectHasNoIsTestProjectProperty">`

src/RunTests/README.md
```diff
@@ -0,0 +1,30 @@
...
+  </ItemGroup>
+</Project>
+```
+This example will include the `Microsoft.Build.RunTests` task for all NuGet-based projects in your repo.
+
+## Example
+To run tests
+```
+$env:MSBUILDENSURESTDOUTFORTASKPROCESSES=1
+msbuild /nodereuse:false /t:MSBuildRunTests
```
nohwnd, line 23: Is `MSBuildRunTests` the final name that user is expected to use, or is this a placeholder. Such long name steals from the appeal for me, because when someone wanted to run the target directly they would use `VSTest` now.

novacole: The name could be something shorter. I am open to any suggestions

src/RunTests/build/Microsoft.Build.RunTests.targets
novacole: I think that makes sense, especially if we're reasonable sure it won't break any existing pipelines

src/RunTests/build/Microsoft.Build.RunTests.targets
novacole: How about `SkipExecution="$(SkipExecution)"` ?

src/RunTests/README.md
novacole: The order would be determined by the user, for example this should work just fine: `msbuild /nodereuse:false /t:Build;Pack;MSBuildRunTests`

src/RunTests/README.md
Evangelink: I think `VSTest` is best as it leaves the door open for different test platform we would want to provide.

src/RunTests/RunTestsTask.cs
novacole: Updated the name of the task and target to use VSTest

src/RunTests/build/Microsoft.Build.RunTests.targets
novacole: Changed to SkipExecution

src/RunTests/build/Microsoft.Build.RunTests.props
novacole: updated to $(MicrosoftBuildPackageVersion)

src/RunTests/README.md
novacole: Updated

src/RunTests/Microsoft.Build.RunTests.csproj
novacole: Are you referring to needing the Microsoft.TestPlatform?

src/RunTests/build/Microsoft.Build.RunVSTest.targets
```diff
@@ -0,0 +1,12 @@
+<Project>
+  <UsingTask TaskName="Microsoft.Build.RunVSTestTask" AssemblyFile="$(MSBuildThisFileDirectory)netstandard2.0\Microsoft.Build.RunVSTest.dll" />
+  <Target Name="RunVSTest" AfterTargets="VSTest" Condition="$(IsTestProject)==true">
```
MarcoRossignoli, line 3: the after here should be `MSBuildRunTests` or?

nohwnd: No, the empty target was renamed to be VSTest, because that is the closest to Test as we can get probably. We should test that this won't break the existing vstest target though.

novacole: I've updated the target name in the msbuild pr to reflect the new "VSTest" name

> the after here should be `MSBuildRunTests` or?

Directory.Packages.props
```diff
@@ -11,19 +11,20 @@
...
-    <PackageVersion Include="MSBuild.ProjectCreation" Version="10.0.0" />
-    <PackageVersion Include="Shouldly" Version="4.2.1" />
-    <PackageVersion Include="xunit" Version="2.6.0" />
-    <PackageVersion Include="xunit.runner.visualstudio" Version="2.4.5" />
+	  <PackageVersion Include="AssemblyShader" Version="1.0.3-preview" />
+	  <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
+	  <PackageVersion Include="MSBuild.ProjectCreation" Version="10.0.0" />
+	  <PackageVersion Include="Shouldly" Version="4.2.1" />
+	  <PackageVersion Include="xunit" Version="2.6.0" />
+	  <PackageVersion Include="xunit.runner.visualstudio" Version="2.4.5" />
```
Evangelink, line 27: These packages are pretty specific to some of your usages. Why not `FluentAssertions`? Why not MSTest or NUnit?

To me you should have something closer to what arcade is doing and condition some of the package deps on the framework.

Shoudly, FluentAssertions... are specific assertion libraries that should also be conditioned by something else. For example, I can use xunit and xunit assert, or xunit and shoudly...

src/RunTests/README.md
```diff
@@ -0,0 +1,39 @@
+# Microsoft.Build.RunVSTest
+
+The `Microsoft.Build.RunVSTest` MSBuild SDK adds support for running tests from MSBuild, similarly to how one would use `dotnet test`.
+
+## Usage in `Directory.Packages.Props`
+In your global.json add the following:
+```json
+{
+  "msbuild-sdks": {
+	"Microsoft.Build.RunVSTest": "1.0.0"
```
dfederm, line 10: You shouldn't have it in both global.json and as a global package reference. You'd use it as one or the other.

In your case I don't see any Sdk.props/targets, so you can remove the global.json section.

src/RunTests/README.md
```diff
@@ -0,0 +1,39 @@
...
+    <GlobalPackageReference Include="Microsoft.Build.RunVSTest" Version="1.0.0" />
+  </ItemGroup>
+</Project>
+```
+This example will include the `Microsoft.Build.RunVSTest` task for all NuGet-based projects in your repo.
+
+## Example
+To run tests
+```
+$env:=1
```
dfederm, line 31: Intentional?

src/RunTests/Microsoft.Build.RunVSTest.csproj
```diff
@@ -0,0 +1,33 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+	<PropertyGroup>
+		<TargetFrameworks>netstandard2.0</TargetFrameworks>
```
dfederm, line 3: There's only 1 so it should be TargetFramework (singular)

src/RunTests/build/Microsoft.Build.RunVSTest.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+	<PropertyGroup>
+		<NugetPath Condition="$(NUGET_PACKAGES)==''">$(USERPROFILE)\.nuget</NugetPath>
```
dfederm, line 3: Isn't there a similar property generated in the *.g.props file you can use?

src/RunTests/build/Microsoft.Build.RunVSTest.props
```diff
@@ -0,0 +1,7 @@
...
+		<NugetPath Condition="$(NUGET_PACKAGES)!=''">$(NUGET_PACKAGES)</NugetPath>
+		<VSTestRunnerVersion Condition="$(VSTestRunnerVersion)==''">$(MicrosoftBuildPackageVersion)</VSTestRunnerVersion>
```
dfederm, line 5: I'm not clear where `MicrosoftBuildPackageVersion` is defined

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,443 @@
...
+                allArgs.Add("--testSessionCorrelationId:" + ArgumentEscaper.HandleEscapeSequenceInArgForProcessStart(VSTestSessionCorrelationId));
+            }
+
+            return allArgs;
+        }
+
+        private Task<int> ExecuteTest()
+        {
+#if NET6_0_OR_GREATER
+            string packagePath = $@"{NugetPath}\packages\microsoft.testplatform\{VSTestRunnerVersion}\tools\net6.0\Common7\IDE\Extensions\TestPlatform\";
```
dfederm, line 424: Seems like this strongly depends on the TF of Microsoft.TestPlatform. What if it changes? Will people need to coordinate this package version with the Microsoft.TestPlatform package one? If so, that should be documented and a table of compatible versions should be published (ie "if you're using Microsoft.TestPlatform version X, use this version of this package")

src/RunTests/build/Microsoft.Build.RunVSTest.props
dfederm: I also am not clear how the Microsoft.Build package version related to the VSTest runner version.

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,443 @@
...
+
+        /// <summary>
+        /// Gets or Sets Runner version of VSTest.
+        /// </summary>
+        public string VSTestRunnerVersion { get; set; }
+
+        /// <summary>
+        /// Gets or Sets Path to nuget package cache.
+        /// </summary>
+        public string NugetPath { get; set; }
```
dfederm, line 160: None of these inputs are marked as required. Should at least some be?

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,443 @@
...
+
+        /// <summary>
+        /// Executes the test. Skips execution if specified.
+        /// </summary>
+        /// <returns>Returns true if the test was executed, otherwise false.</returns>
+        public override bool Execute()
+        {
+            var traceEnabledValue = Environment.GetEnvironmentVariable("VSTEST_BUILD_TRACE");
+            if (SkipExecution)
```
dfederm, line 169: Should this be skipped MSBuild on the Target instead of invoking the task just to short-circuit immediately?

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,443 @@
...
+            }
+            else
+            {
+                allArgs.Add(ArgumentEscaper.HandleEscapeSequenceInArgForProcessStart(TestFileFullPath));
+            }
+
+            // Console logger was not specified by user, but verbosity was, hence add default console logger with verbosity as specified
+            if (!string.IsNullOrEmpty(VSTestVerbosity) && !isConsoleLoggerSpecifiedByUser)
+            {
+                var normalTestLogging = new List<string>() { "n", "normal", "d", "detailed", "diag", "diagnostic" };
```
dfederm, line 305: If you're only doing `Contains` operations, these should be `HashSet<string>` instead.

Also, these should be made `static` to avoid allocating each time the task runs

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,443 @@
...
+            {
+                allArgs.Add("--testSessionCorrelationId:" + ArgumentEscaper.HandleEscapeSequenceInArgForProcessStart(VSTestSessionCorrelationId));
+            }
+
+            return allArgs;
+        }
+
+        private Task<int> ExecuteTest()
+        {
+#if NET6_0_OR_GREATER
```
dfederm, line 423: You're only targetting netstandard so this ifdef is never going to be used

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,443 @@
...
+
+            // Avoid logging "Task returned false but did not log an error." on test failure, because we don't
+            // write MSBuild error. https://github.com/dotnet/msbuild/blob/51a1071f8871e0c93afbaf1b2ac2c9e59c7b6491/src/Framework/IBuildEngine7.cs#L12
+            var allowfailureWithoutError = BuildEngine.GetType().GetProperty("AllowFailureWithoutError");
+            allowfailureWithoutError?.SetValue(BuildEngine, true);
+
+            return ExecuteTest().GetAwaiter().GetResult() == 0;
+        }
+
+        internal IEnumerable<string> CreateArgument()
```
dfederm, line 203: Naming nit: Argument*s* (plural)

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,443 @@
...
+                    Thread.Sleep(1000);
+                }
+
+                Debugger.Break();
+            }
+
+            // Avoid logging "Task returned false but did not log an error." on test failure, because we don't
+            // write MSBuild error. https://github.com/dotnet/msbuild/blob/51a1071f8871e0c93afbaf1b2ac2c9e59c7b6491/src/Framework/IBuildEngine7.cs#L12
+            var allowfailureWithoutError = BuildEngine.GetType().GetProperty("AllowFailureWithoutError");
+            allowfailureWithoutError?.SetValue(BuildEngine, true);
```
dfederm, line 198: This is a bit of a hack. Should an error just be logged to give the user some indication of what's happening?

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,443 @@
...
+                FileName = $"{packagePath}vstest.console.exe",
+                Arguments = string.Join(" ", CreateArgument()),
+                UseShellExecute = false,
+            };
+
+            using (var activeProcess = new Process { StartInfo = processInfo })
+            {
+                activeProcess.Start();
+                activeProcess.WaitForExit();
+                return Task.FromResult(activeProcess.ExitCode);
```
dfederm, line 439: Since you're using `Task.FromResult`, no need to make this method return a task at all

src/RunTests/build/Microsoft.Build.RunVSTest.targets
```diff
@@ -0,0 +1,12 @@
...
+	  <RunVSTestTask VSTestRunnerVersion="$(VSTestRunnerVersion)" TestFileFullPath="$(TargetPath)" VSTestSetting="$(VSTestSetting)" VSTestTestAdapterPath="$(VSTestTestAdapterPath)"
```
dfederm, line 4: Style nit: add newlines between all arguments for readability

dfederm:
You'll also want to update the Traversal SDK to recurse the `Test` target I think.

Does it just work for slns by executing the target ("Test") on every project in the sln?

src/RunTests/build/Microsoft.Build.RunVSTest.targets
novacole: Updated name to "Test", now that the target is opt in

src/RunTests/README.md
novacole: Updated to show usage for both Sdk and package reference

src/RunTests/build/Microsoft.Build.RunVSTest.props
novacole: Updated this default value to be the most recent release non-preview version. However, presumably customers would specify the version they would like to use.

novacole:
> ~You'll also want to update the Traversal SDK to recurse the `Test` target I think.~
> 
> EDIT: Disregard. It's already here: https://github.com/microsoft/MSBuildSdks/blob/b54b4a19dd55281bd87e2f8dd482d444c44dc666/src/Traversal/Sdk/Traversal.targets#L181C19-L181C19. Probably for `dotnet test` support already.
> 
> Does it just work for slns by executing the target ("Test") on every project in the sln?

Yes

Directory.Packages.props
novacole: These are not included in the package.

src/RunTests/README.md
```diff
@@ -0,0 +1,82 @@
...
+  <ItemGroup>
+    <PackageVersion Include="Microsoft.Build.RunVSTest" Version="1.0.0" />
+  </ItemGroup>
+</Project>
+```
+in your .csproj file
+```
+<Project>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Build.RunVSTest" Version="1.0.0" />
```
johnterickson, line 35: Hmm with this I am getting version conflicts - which doesn't seem right.

```
Z:\Mycroft\Test\Mycroft.Tests\Mycroft.Tests.csproj : error NU1605: Warning As Error: Detected package downgrade: System.Collections.Immutable from 7.0.0 to 5.0.0. Reference the package directly from the project to select a different version.  [Z:\Mycroft\dirs.proj]
Z:\Mycroft\Test\Mycroft.Tests\Mycroft.Tests.csproj : error NU1605:  Mycroft.Tests -> Microsoft.Build.RunVSTest 1.0.296-gf180ca4013 -> Microsoft.Build.Tasks.Core 17.8.3 -> System.Collections.Immutable (>= 7.0.0)  [Z:\Mycroft\dirs.proj]
Z:\Mycroft\Test\Mycroft.Tests\Mycroft.Tests.csproj : error NU1605:  Mycroft.Tests -> System.Collections.Immutable (>= 5.0.0) [Z:\Mycroft\dirs.proj]
```

src/RunTests/Microsoft.Build.RunVSTest.csproj
```diff
@@ -0,0 +1,34 @@
...
+		<Description>Runs VS Test Console</Description>
+		<NoDefaultExcludes>true</NoDefaultExcludes>
+		<BuildOutputTargetFolder>build\</BuildOutputTargetFolder>
+		<AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
+		<DevelopmentDependency>true</DevelopmentDependency>
+		<LangVersion>latest</LangVersion>
+		<SymbolPackageFormat>snupkg</SymbolPackageFormat>
+	</PropertyGroup>
+	<ItemGroup>
+		<PackageReference Include="Microsoft.Build.Tasks.Core" />
```
johnterickson, line 15: These should both have `PrivateAssets="All"`

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,434 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+
+namespace Microsoft.Build
+{
+    /// <summary>
+    /// Runs tests with vstest.
+    /// </summary>
+    public class RunVSTestTask : Task
```
rainersigwald, line 18: Since this is launching an executable, inheriting from `ToolTask` may let you rely on common implementation for stuff like reading from stdin/out and other process-launch minutia and simplify the new code here.

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,434 @@
...
+                FileName = $"{packagePath}vstest.console.exe",
+                Arguments = string.Join(" ", CreateArguments()),
+                UseShellExecute = false,
+                RedirectStandardError = true,
+                RedirectStandardOutput = true,
+            };
+
+            using var activeProcess = new Process { StartInfo = processInfo };
+            activeProcess.Start();
+            using StreamReader errReader = activeProcess.StandardError;
```
johnterickson, line 424: This deadlocks for me.  Moving to ToolTask should fix things

src/RunTests/RunVSTestTask.cs
MarcoRossignoli: @nohwnd can you give some drive here as you worked on the tool task some time ago?

src/RunTests/README.md
```diff
@@ -0,0 +1,82 @@
...
+```xml
+<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Sdk Name="Microsoft.Build.RunVSTest"/>
+  ...
+</Project>
+```
+
+## For projects that use packages references. In your `Directory.Packages.props`:
+```xml
+<Project>
```
MarcoRossignoli, line 24: are we missing here the `<UseMSBuildTestInfrastructure>true</UseMSBuildTestInfrastructure>` ?

MarcoRossignoli: So the non tests project will import the empty `Test` target and won't fail for target not found for run at solution level?

novacole: > <UseMSBuildTestInfrastructure>true</UseMSBuildTestInfrastructure>

updated readme to include

src/RunTests/RunVSTestTask.cs
novacole: Updated to be a tooltask

src/RunTests/RunVSTestTask.cs
novacole: updated to be a tooltask now

src/RunTests/README.md
```diff
@@ -0,0 +1,88 @@
...
+  <ItemGroup>
+    <PackageVersion Include="Microsoft.Build.RunVSTest" Version="1.0.0" />
+  </ItemGroup>
+</Project>
+```
+in your .csproj file opt-in to using 'UseMSBuildTestInfrastructure' and add the package reference to your test project csproj files
+```
+<Project>
+  <PropertyGroup>
+	<UseMSBuildTestInfrastructure>true</UseMSBuildTestInfrastructure>
```
dfederm, line 35: Can the package just set this in the props file?

src/RunTests/README.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+## Sln
+For sln project it is sufficent to simply opt-in to using 'UseMSBuildTestInfrastructure' and add the package reference to your test project csproj files
+```
+<Project>
+<PropertyGroup>
+	<UseMSBuildTestInfrastructure>true</UseMSBuildTestInfrastructure>
+</PropertyGroup>
+<UseMSBuildTestInfrastructure>true</UseMSBuildTestInfrastructure>
```
dfederm, line 65: This line looks like it needs to go

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,376 @@
...
+
+        protected override MessageImportance StandardErrorLoggingImportance => MessageImportance.High;
+
+        /// <inheritdoc/>
+        protected override string GenerateFullPathToTool()
+        {
+#if NET6_0_OR_GREATER
+            return $@"{NugetPath}microsoft.testplatform\{VSTestRunnerVersion}\tools\net6.0\Common7\IDE\Extensions\TestPlatform\vstest.console.exe";
+#else
+            return $@"{NugetPath}microsoft.testplatform\{VSTestRunnerVersion}\tools\net462\Common7\IDE\Extensions\TestPlatform\vstest.console.exe";
```
dfederm, line 175: Seems to me like we should just mimic what `dotnet test` does for the `dotnet` scenario, and for the VS scenario just find vstest from the VS install. Digging through the package cache seems like a pretty brittle thing to do, especially since the versions will need to align with some PackageReference for it to work consistently.

dfederm: In fact, for the `dotnet` case can we actually just invoke the target which `dotnet test` does?

dfederm: In `C:\Program Files\dotnet\sdk\...\Microsoft.TestPlatform.targets` I see:

```
<Target Name="VSTest" DependsOnTargets="ShowInfoMessageIfProjectHasNoIsTestProjectProperty">
```

So we should probably just have something like this:
```xml
<Target Name="RunVSTestAfterTest" AfterTargets="Test" DependsOnTargets="VSTest" />
```

And then in the non-`dotnet` case, figure out how to import `Microsoft.TestPlatform.targets` or something. This would probably only work for SDK-style projects, but that's probably OK?

That way we never are maintaining how VSTest is invoked at all and it's always consistent with `dotnet test`.

novacole: > And then in the non-dotnet case, figure out how to import Microsoft.TestPlatform.targets or something. 
Microsoft.TestPlatform.targets invokes a task that invokes dotnet

src/RunTests/build/Microsoft.Build.RunVSTest.props
```diff
@@ -0,0 +1,14 @@
...
+<!--
+  Copyright (c) Microsoft Corporation. All rights reserved.
+  
+  Licensed under the MIT license.
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <NugetPath Condition="$(NugetPath)==''">$(NuGetPackageRoot)</NugetPath>
+    <NugetPath Condition="$(NugetPath)=='' And $(NUGET_PACKAGES)!=''">$(NUGET_PACKAGES)</NugetPath>
+    <SkipExecution Condition="$(SkipExecution)==''">false</SkipExecution>
```
dfederm, line 11: `SkipExecution` is pretty vague. `SkipTestExecution` maybe?

Also, why is this needed? I'm not sure I get the scenario of `msbuild /t:Build;Test /p:SkipExecution=true`

src/RunTests/RunVSTestTask.cs
```diff
@@ -0,0 +1,376 @@
...
+
+        protected override MessageImportance StandardOutputLoggingImportance => MessageImportance.High;
+
+        protected override MessageImportance StandardErrorLoggingImportance => MessageImportance.High;
+
+        /// <inheritdoc/>
+        protected override string GenerateFullPathToTool()
+        {
+#if NET6_0_OR_GREATER
+            return $@"{NugetPath}microsoft.testplatform\{VSTestRunnerVersion}\tools\net6.0\Common7\IDE\Extensions\TestPlatform\vstest.console.exe";
```
dfederm, line 173: The net6.0 directory doesn't exist in the Microsoft.TestPlatform package:

image

dfederm (approved):

nohwnd:
I have been doing quite a lot of changes to vstest targets and tasks, I would love this to take dependency on stable version of https://www.nuget.org/packages/Microsoft.TestPlatform.Build which ships them.

novacole:
@microsoft-github-policy-service agree company="Microsoft"

nohwnd:
Microsoft.TestPlatform**.Build** contains the msbuild task and target that we use to invoke vstest.console, depending I think that depending on it could avoid duplicating the task and target in this sdk, e.g. src/RunTests/RunVSTestTask.cs, we also have a new task and target that integrates better with msbuild and terminal logger that we are actively developing.

Issue #155, opened 2020-03-12
Microsoft.Build.NoTargets does not work with multi targeting when project extension ist .msbuildproj


bash:
## Minimal, Reproducible Example
Source: https://github.com/bash/MSBuildNoTargetsIssueRepro 
Failing Travis build: https://travis-ci.com/github/bash/MSBuildNoTargetsIssueRepro/builds/152979950/152979792
Error Message:

> /usr/share/dotnet/sdk/3.1.102/Microsoft.Common.CurrentVersion.targets(1175,5): error MSB3644: The reference assemblies for .NETFramework,Version=v4.0 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks [/home/travis/build/bash/MSBuildNoTargetsIssueRepro/MSBuildNoTargetsIssueRepro.msbuildproj]

## Possible Workaround

I believe the issue is that Microsoft.Common.CrossTargeting.targets is never imported when the project extension doesn't correspond to a language.
My [workaround](https://github.com/bash/MSBuildNoTargetsIssueRepro/tree/workaround) sets the property`LanguageTargets` to a custom file, that imports CrossTargeting.targets

ViktorHofer:
Another easier workaround, use the .csproj extension instead. Not perfect but it works...

jzabroski:
@bash I think you have a typo in the issue. ist -> is

@ViktorHofer I get a build error when using csproj now that I upgraded to VS2022 and uninstalled all old SDKs.

Severity	Code	Description	Project	File	Line	Suppression State
Error	MSB3644	The reference assemblies for .NETFramework,Version=v4.0 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks	PowerShellBuildProject	C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets	1220	

My xml is

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project Sdk="Microsoft.Build.NoTargets/3.3.0" InitialTargets="RunBeforeAnySsrsReport">
  <!-- ReSharper disable UnknownProperty -->
  <PropertyGroup>
[...]

PR #478, opened 2023-09-08
Disable NuGet pack for NoTargets projects

MattKotsenas:
# Scenario

I have a large solution / traversal project that includes projects that create NuGet packages, and also NoTargets projects to hold other targets. As part of my CI/CD pipeline I want to run `dotnet pack` on the top-level project to create all NuGet packages.

Note that I use the `.csproj` extension for NoTargets projects to avoid the issues mentioned in #155 .

# The problem

NoTargets projects fail with the following error message:

> C:\Program Files\dotnet\sdk\7.0.400\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(221,5):
> error NU5017: Cannot create a package that has no dependencies nor content.
> [/path/to/my/NoTargets.csproj]

# Workarounds

Adding `<IsPackable>false</IsPackable>` to every NoTargets project

# The fix

Update the NoTargets SDK to set IsPackable to false by default, and add unit tests to validate behavior.


jeffkl:
@MattKotsenas This would be a breaking change for users that are packing NoTargets projects today.  For your repository, could you place something like this in your Directory.Build.targets:

```xml
<Project>
  <PropertyGroup>
    <IsPackable Condition="'$(UsingMicrosoftNoTargetsSdk)' == 'true'">false</IsPackable>
  </PropertyGroup>
</Project>
```

I would be fine with a new property that you could set in your Directory.Build.props:

```xml
<Project>
  <PropertyGroup>
    <NoTargetsIsPackable>false</NoTargetsIsPackable>
  </PropertyGroup>
</Project>
```

So if `NoTargetsIsPackable` was set to `false`, then the NoTargets build logic would set `IsPackable` to false for you.  But I think making it not packable by default is going to break people.

jeffkl:
@MattKotsenas thanks anyway for the contribution.  I think in this case avoiding a breaking change is the top priority, we both found instances where people are using NoTargets to create packages.  At least with an entry in your `Directory.Build.targets` you can disable the functionality for a whole repository.  I'd be fine with the new property that disables it as well.

PR #484, opened 2023-09-25
Remove old TFMs from Artifacts SDK

AndyGerlicher:
Fixes an issue with copy on write changes and the wrong dll getting used (`net472` dll would call into code that was disabled by ifdef). Turned out to be easier to just remove the legacy TFM and only target `netstandard2.0`.


jeffkl (commented):
Let's rev the `version.json` as well to the next major version so people know its a "breaking change".

jeffkl (approved):

erikmav (approved):

PR #489, opened 2023-10-06
Rename the RobocopyFiles task

erikmav:
To enable migrating old internal code that uses the same name without having packages override one another.


jeffkl:
The CI is flaky because MSBuild calls are throwing for some reason, we can ignore those.

I'm pretty sure we re-used the same target name so when users migrated from CoreXT to retail, any custom build logic they had to run before/after robocopy would continue to work.  Can you elaborate more on how migrating can cause conflicts?  Users should be on CoreXT or retail but not both right?

erikmav:
Unneeded after all - just needed to deal with old spaghetti code better.

PR #491, opened 2023-10-16
Make System.Threading.Tasks.Dataflow dependency private

### LINKED ISSUE #490: Microsoft.Build.Artifacts package should not depend on System.Threading.Tasks.Dataflow
The Microsoft.Build.Artifacts package currently has a dependency on System.Threading.Tasks.Dataflow:

image

This package dependency means that a copy of `System.Threading.Tasks.Dataflow.dll` is copied into the output of projects using Microsoft.Build.Artifacts.

From what I can tell, this dependency is a mistake. While `Microsoft.Build.Artifacts.dll` does have a reference to `System.Threading.Tasks.Dataflow.dll`, the assembly is already shipped inside the package:

image

There should be no reason to have the package dependency. It would be great to get it removed!

bording:
Fixes #490


erikmav (approved):

jeffkl (approved):
Thanks for the contribution!

PR #507, opened 2023-11-20
Fix typo in Traversal README

fl3pp:
The example in the `REAME.md` of the Traversal SDK has a typo that causes the example to fail.


jeffkl (approved):
Thank you for fixing this!

PR #510, opened 2023-12-06
Fix for self-copy deleting the source file in CoW copying

erikmav:
In CoW Copy and Artifacts, do not copy when the source and destination paths are the same (canonicalized) to avoid deleting the source file when a copy-on-self is happening. This does not handle the case of the same file via differing symlinked paths.

### DIFF
--- a/src/Artifacts.UnitTests/RobocopyTests.cs
+++ b/src/Artifacts.UnitTests/RobocopyTests.cs
@@ -351,6 +351,45 @@ namespace Microsoft.Build.Artifacts.UnitTests
         }
 
         [Fact]
+        public void ExplicitSelfCopiesShouldNoOp()
+        {
+            // Use a relative path to allow aliasing a local file and its full path.
+            string localSource = Path.Combine(Environment.CurrentDirectory, "foo.txt");
+            File.WriteAllText(localSource, string.Empty);
+
+            BuildEngine buildEngine = BuildEngine.Create();
+            MockFileSystem fs = new MockFileSystem();
+
+            Robocopy copyArtifacts = new Robocopy
+            {
+                BuildEngine = buildEngine,
+                Sources = new ITaskItem[]
+                {
+                    new MockTaskItem(localSource)
+                    {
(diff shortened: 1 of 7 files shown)


erikmav:
An internal repo is hitting a problem with the 1.0.282 logic that added in the off-by-default "delete before copy" logic gated in regular MSBuild Copy by feature flag check `if (ChangeWaves.AreFeaturesEnabled(ChangeWaves.Wave17_8)`. @KirillOsenkov FYI as you had filed an item against this Copy task to align with MSBuild logic, which I assume no one in the internal repo has tried yet. Also @AndyGerlicher @jeffkl FYI

erikmav:
Still an open TODO in the new logic related to true paths and symlinked/junctioned dirs. I don't have a good solution for this yet. An alternative to this change is to revert the destination delete logic back to before the feature flag was added in MSBuild Copy. In the interim getting this change in and a new package spun up will help with migrating the repo I'm focused on.

erikmav:
Just to clarify - if the MSBuild Copy feature flag that deletes files before copying is turned on, any repo that has a copy-on-self like this large repo does will end up deleting the source file and failing. Some sort of fix similar to this - or not activating the feature flag - will be needed to avoid this problem.

The CoW Copy code, though derived from MSBuild's Copy, has the disadvantage that `File.Copy` will silently ignore copy-onto-self cases, both at the path level and via directory symlinks/junctions. The CoW code hence has a problem in that it has to replicate similar checks without relying on checks in the Win32 subsystem that check for the same file via comparing at the file handle level.

erikmav:
Proceeding to commit and publish new packages to unblock testing.

KirillOsenkov:
We have been using 1.0.282 from the day it shipped and we haven't noticed anything wrong. @inthemedium to confirm.

The changes you made today don't look breaking to us. Did you expect us to be impacted in any way?

erikmav:
Yes there is a problem there. Working on a minimal repro of the various cases then will fix a bug and link it here.

PR #536, opened 2024-03-05
Update official build pipeline

AndyGerlicher:


dfederm (approved):

jeffkl (approved):
You might need to manually kick off a build with official signing to test that it works, Real signing only happens for tagged commits

PR #559, opened 2024-04-20
Set default SlnGenProjectName for Traversal projects

MattKotsenas:
microsoft/slngen#557 added support for customizing the solution name using the MSBuild property `<SlnGenProjectName>`.

Today, if a user runs SlnGen with default properties, every VS window will unhelpfully be named "dirs". With this change, if a user hasn't set this property, this change sets it to the parent directory name, which is _usually_ more descriptive. In some cases that may result in equally unhelpful names like "src", but that's no worse than today.

This may be considered a breaking change for SlnGen users in some scenarios. Specifically, if they were running slngen with no parameters set and then assuming the resulting solution would be named `{dirs}.sln`. In many cases, anyone that's writing scripts with slngen is also setting `-o|--solutionfile`, so they wouldn't be impacted, but I'm sure _someone_ is relying on the name.

### DIFF
--- a/src/Traversal.UnitTests/TraversalTests.cs
+++ b/src/Traversal.UnitTests/TraversalTests.cs
@@ -208,10 +208,12 @@ namespace Microsoft.Build.Traversal.UnitTests
         [InlineData("TraversalProjectNames", "custom.proj", "custom.proj")]
         [InlineData("TraversalProjectNames", null, "dirs.proj")]
         [InlineData("UsingMicrosoftTraversalSdk", null, "true")]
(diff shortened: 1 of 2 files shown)


MattKotsenas:
Open to thoughts / considerations on the benefits / risks here, but wanted to open a PR to start the conversation :)

MattKotsenas:
I haven't really needed this since opening it, and given that it's likely to be a breaking change for some folks, I'm going to close for now. We can revisit if anyone clamors for it.

PR #563, opened 2024-06-27
Dont use IndexOf as that has Cuture issues

hknielsen:
Theres currently some culture issues using IndexOf as mentioned in this issue https://github.com/dotnet/msbuild/issues/5502.
https://github.com/dotnet/msbuild/issues/5502 fixes it, but its only enabled in a late wave, that we are not depending on yet.
This PR uses contains instead, it seem that fixes the issue with Culture for now.

### DIFF
--- a/src/Traversal/Sdk/Traversal.props
+++ b/src/Traversal/Sdk/Traversal.props
@@ -13,7 +13,7 @@
     -->
     <TraversalProjectNames Condition=" '$(TraversalProjectNames)' == '' ">dirs.proj</TraversalProjectNames>
 
-    <IsTraversal Condition=" '$(IsTraversal)' == '' And $(TraversalProjectNames.IndexOf($(MSBuildProjectFile), System.StringComparison.OrdinalIgnoreCase)) >= 0 ">true</IsTraversal>
+    <IsTraversal Condition=" '$(IsTraversal)' == '' And '$(TraversalProjectNames.IndexOf($(MSBuildProjectFile), System.StringComparison.OrdinalIgnoreCase))' != '0' ">true</IsTraversal>
 
     <!--
       NuGet should always restore Traversal projects with "PackageReference" style restore.  Setting this property will cause the right thing to happen even if there aren't any PackageReference items in the project.


hknielsen:
Ah Contains dont work in all MSBuild versions, ie the one is used in CI here.
@jeffkl do you have any good ideas? What is the `IsTraversal` property used for? AFAICT its set after ie. Directory.Build.props is imported, so users don't use it to conditionally do logic in those.

jeffkl:
Its easy for tooling to differentiate between a solution file and a project because they have different extensions.  However, its much harder if two projects end in `.proj`.  The idea behind this property is so that tooling can treat a project different if its really just a traversal project.  I don't think a ton of tooling has a dependency on it but we certainly can't get rid of the property.  I know I used it in [SlnGen](https://github.com/search?q=repo%3Amicrosoft%2Fslngen%20istraversal&type=code). 

It is set after `Directory.Build.props` but before the project and `Directory.Build.targets`.  Most tools just evaluate a project and then inspect the properties.

From the issue you linked, it looks more like not properly converting the `0` to a string: https://github.com/dotnet/msbuild/issues/9757

Are you seeing issues because of this call to `IndexOf()`?

hknielsen:
> From the issue you linked, it looks more like not properly converting the 0 to a string: https://github.com/dotnet/msbuild/issues/9757
> 
> Are you seeing issues because of this call to IndexOf()?

Yes in this case its because of the call to `IndexOf` - we could also make the Property not use that, as its traversal by nature when this props file is imported?

jeffkl:
I'm a little nervous about changing behavior.  Would comparing a string work better?
```diff
- <IsTraversal Condition=" '$(IsTraversal)' == '' And $(TraversalProjectNames.IndexOf($(MSBuildProjectFile), System.StringComparison.OrdinalIgnoreCase)) >= 0 ">true</IsTraversal>
+ <IsTraversal Condition=" '$(IsTraversal)' == '' And '$(TraversalProjectNames.IndexOf($(MSBuildProjectFile), System.StringComparison.OrdinalIgnoreCase))' != '0' ">true</IsTraversal>
```

hknielsen:
@jeffkl - good idea, that seem to work as well, so im changing the code to do that instead

src/Traversal/Sdk/Traversal.props
```diff
@@ -13,7 +13,7 @@
     -->
     <TraversalProjectNames Condition=" '$(TraversalProjectNames)' == '' ">dirs.proj</TraversalProjectNames>
 
-    <IsTraversal Condition=" '$(IsTraversal)' == '' And $(TraversalProjectNames.IndexOf($(MSBuildProjectFile), System.StringComparison.OrdinalIgnoreCase)) >= 0 ">true</IsTraversal>
+    <IsTraversal Condition=" '$(IsTraversal)' == '' And $(TraversalProjectNames.Contains($(MSBuildProjectFile), System.StringComparison.OrdinalIgnoreCase)) ">true</IsTraversal>
```
hknielsen, line 16:
```suggestion
    <IsTraversal Condition=" '$(IsTraversal)' == '' And '$(TraversalProjectNames.IndexOf($(MSBuildProjectFile), System.StringComparison.OrdinalIgnoreCase))' != '0' ">true</IsTraversal>


```

dfederm:
Closing stale PR

PR #567, opened 2024-09-09
Better Artifact defaults when using built-in

AndyGerlicher:


dfederm:
My preferred change would be to change:
 
```
<EnableArtifacts Condition="'$(UseArtifactsOutput)' == 'true'">false</EnableArtifacts>
```

To:
```
<EnableDefaultArtifacts Condition="'$(UseArtifactsOutput)' == 'true'">false</EnableDefaultArtifacts>
```

and I guess move it from Microsoft.Build.Artifacts.Common.props to Microsoft.Build.Artifacts.props.

The "artifacts output layout" really just redirects the default outputs to a repo-rooted location. It doesn't solve the problem of arbitrary copies. So IMO only the *default* artifacts should be disabled, ie the implicit `<Artifact Include="$(OutputPath)" DestinationFolder="$(ArtifactsPath)" />`. Any explicit `Artifact` or `Robocopy` items should still be respected.

dfederm (approved):

dfederm (approved):

PR #570, opened 2024-09-24
Fix NoTargetsTests.cs

stan-sz:


AndyGerlicher:
I pulled this fix into #573. Thanks!

PR #571, opened 2024-09-24
Remove old TFM support

AndyGerlicher:


azure-pipelines.yml
```diff
@@ -5,8 +5,6 @@ variables:
   ArtifactsDirectoryName: 'artifacts'
   BuildConfiguration: 'Debug'
   BuildPlatform: 'Any CPU'
-  DotNet6Version: '6.x'
-  DotNet7Version: '7.x'
```
erikmav, line 9: > Do [](http://example.com/codeflow?start=1&length=3)

Add 9.x?

erikmav (approved):
:shipit:

Directory.Packages.props
```diff
@@ -3,7 +3,7 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <MicrosoftBuildPackageVersion Condition="'$(TargetFramework)' == 'netstandard2.0'">16.9.0</MicrosoftBuildPackageVersion>
     <MicrosoftBuildPackageVersion Condition="'$(TargetFramework)' == 'net46'">15.9.20</MicrosoftBuildPackageVersion>
-    <MicrosoftBuildPackageVersion Condition="'$(TargetFramework)' == 'net7.0'">17.7.2</MicrosoftBuildPackageVersion>
+    <MicrosoftBuildPackageVersion Condition="'$(TargetFramework)' == 'net8.0'">17.7.2</MicrosoftBuildPackageVersion>
```
dfederm, line 6: I wonder if we can use a newer MSBuild as well. The main purpose of this condition is because MSBuild also bumps its TF, so we have to stick to the newest that supports a given TF

dfederm: Yea, looks like we can just use the latest (17.11.4).

You can actually just remove this line entirely and let net8 fall down to the line below. And then optionally update the version in the line below as well

PR #581, opened 2024-10-14
Update to latest System.Text.Json with security fix

erikmav:
Fix build breaks from Component Governance alert.


AndyGerlicher (approved):

rainersigwald:
Note that this will break the package on older Visual Studio versions that don't have STJ 8.0.5 yet, which is everything less than 17.12 GA (which isn't public yet).

PR #599, opened 2025-01-09
Add Cargo Build Sdk

novacole:
This sdk adds support for building rust projects with cargo within MSBuild.


src/CargoBuild/Microsoft.Build.CargoBuild.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+	<PropertyGroup>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<Nullable>enable</Nullable>
+		<TargetFrameworks>net6.0;net472</TargetFrameworks>
+		<GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+		<PackageVersion>1.0.0</PackageVersion>
```
jeffkl, line 7: You'll want to use Nerdbank.GitVersioning in order to have the version generated with a deterministic value, so remove this and add a `version.json` like the other SDKs.

https://github.com/microsoft/MSBuildSdks/blob/main/src/Traversal/version.json

src/CargoBuild/README.md
```diff
@@ -0,0 +1,91 @@
...
+1) in your global.json
+```json
+ "msbuild-sdks": {
+    ...,
+    "Microsoft.Build.CargoBuild": "1.0.270-gf406f8eaa0"
+  },
+```
+
+2) foreach rust project a .cargosproj project file at the same level as your cargo.toml file. The project file should import the CargoBuild sdk.
+```xml
```
jeffkl, line 15: nit: In this XML blob I see 4 space indenting and the one below it is 2 space indenting.  I think XML should be 2 space indenting personally.

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,93 @@
+<!--
+  Copyright (c) Microsoft Corporation. All rights reserved.
+  
+  Licensed under the MIT license.
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+	<PropertyGroup>
+		<UsingMicrosoftCargoBuildSdk>true</UsingMicrosoftCargoBuildSdk>
```
AndyGerlicher, line 9: Is this used anywhere? Seems only read in the unit test. If it is used, should probably set this to true on Condition it's empty.

src/CargoBuild/Microsoft.Build.CargoBuild.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+	<PropertyGroup>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<Nullable>enable</Nullable>
+		<TargetFrameworks>net6.0;net472</TargetFrameworks>
```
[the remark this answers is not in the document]
novacole: Yes, it was because of the lack of a netstandard version of `Microsoft.Build`

src/CargoBuild/sdk/Sdk.props
dfederm: This is a common pattern to use as a signal that the SDK was used. The .NET SDK does the same thing, and we do for NoTargets, etc as well

dfederm: Example from .NET SDK: https://github.com/dotnet/sdk/blob/41bcaa914b54c96ad20e784d0baeccf2d7ab7fa1/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props#L22-L31

And Microsoft.NET.Sdk.Web:
https://github.com/dotnet/sdk/blob/41bcaa914b54c96ad20e784d0baeccf2d7ab7fa1/src/WebSdk/Web/Targets/Sdk.Server.props#L17

And then our NoTargets:
https://github.com/microsoft/MSBuildSdks/blob/a72501989d5b03f3f22257d41416d59695f1bd3c/src/NoTargets/Sdk/Sdk.props#L10

So IMHO, leave this property as you have it.

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+		<MSBuildAllProjects Condition="'$(MSBuildToolsVersion)' != 'Current'">$(MSBuildAllProjects);$(MsBuildThisFileFullPath)</MSBuildAllProjects>
+	</PropertyGroup>
+
+	<PropertyGroup>
+		<!-- Copy logic to know if managed targets got imported: https://github.com/dotnet/sdk/blob/49002c14cf91ecd08e79d6184dbd4716c005b509/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.targets#L25-L27 -->
```
dfederm, line 14: Can the comment explain why?

So, is it expected that a cargobuild project would be named "csproj"?

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+		<ManagedLanguageTargetsGotImported Condition="'$(MSBuildProjectExtension)' == '.csproj'">true</ManagedLanguageTargetsGotImported>
+		<ManagedLanguageTargetsGotImported Condition="'$(MSBuildProjectExtension)' == '.vbproj'">true</ManagedLanguageTargetsGotImported>
+		<ManagedLanguageTargetsGotImported Condition="'$(MSBuildProjectExtension)' == '.fsproj'">true</ManagedLanguageTargetsGotImported>
+	</PropertyGroup>
+
+	<Import Project="$(CustomBeforeNoTargetsProps)" Condition=" '$(CustomBeforeNoTargetsProps)' != '' And Exists('$(CustomBeforeNoTargetsProps)') " />
+
+	<PropertyGroup>
+		<!-- Disable default Compile and EmbeddedResource items for NoTargets projects -->
```
dfederm, line 23: Nit: "NoTargets" -> "CargoBuild"

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+		<EnableDefaultCompileItems>false</EnableDefaultCompileItems>
```
dfederm, line 24: Do you not want `**\*.rs` files to be Compile items?

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+		<EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
+
+		<!--
+      NuGet should always restore Traversal projects with "PackageReference" style restore.  Setting this property will cause the right thing to happen even if there aren't any PackageReference items in the project.
```
dfederm, line 28: Fix comment, should mention that this is a traversal project

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+    -->
+		<RestoreProjectStyle>PackageReference</RestoreProjectStyle>
+
+		<!-- Targeting packs shouldn't be referenced as traversal projects don't compile. -->
```
dfederm, line 32: Remove reference to traversal project

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+		<!-- Don't generate assembly info -->
+		<GenerateAssemblyInfo Condition="'$(GenerateAssemblyInfo)' == ''">false</GenerateAssemblyInfo>
+
+		<!-- Don't generate editor config file -->
+		<GenerateMSBuildEditorConfigFile Condition="'$(GenerateMSBuildEditorConfigFile)' == ''">false</GenerateMSBuildEditorConfigFile>
+
+		<!-- Don't log the high priority message mentioning this project's name (or copy the product we didn't build). -->
+		<SkipCopyBuildProduct Condition="'$(SkipCopyBuildProduct)' == ''">true</SkipCopyBuildProduct>
+
+		<!-- Don't automatically reference assembly packages since NoTargets don't need reference assemblies -->
```
dfederm, line 51: Don't mention NoTargets

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+		<AutomaticallyUseReferenceAssemblyPackages Condition="'$(AutomaticallyUseReferenceAssemblyPackages)' == ''">false</AutomaticallyUseReferenceAssemblyPackages>
+		<NoCompilerStandardLib Condition="'$(NoCompilerStandardLib)' == ''">false</NoCompilerStandardLib>
+		<NoStdLib Condition="'$(NoStdLib)' == ''">true</NoStdLib>
+
+		<!-- Disable Visual Studio's Fast Up-to-date Check and rely on MSBuild to determine -->
+		<DisableFastUpToDateCheck Condition="'$(DisableFastUpToDateCheck)' == ''">true</DisableFastUpToDateCheck>
+	</PropertyGroup>
+
+	<ItemDefinitionGroup Condition=" '$(NoTargetsDoNotReferenceOutputAssemblies)' != 'false' ">
```
dfederm, line 60: Bad copy/pasted property name.

dfederm: I'm actually not convinced that we need the condition at all? Since you never would want to reference the output assembly in the normal way.

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+			<ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+			<SkipGetTargetFrameworkProperties>true</SkipGetTargetFrameworkProperties>
+		</ProjectReference>
+	</ItemDefinitionGroup>
+
+	<Target Name="CreateManifestResourceNames" />
+
+	<Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" Condition=" '$(MicrosoftCommonPropsHasBeenImported)' != 'true' "/>
+
+	<Import Project="$(CustomAfterNoTargetsProps)" Condition=" '$(CustomAfterNoTargetsProps)' != '' And Exists('$(CustomAfterNoTargetsProps)') " />
```
dfederm, line 75: Fix property name

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+
+	<!-- For CPS/VS support. Importing in .props allows any subsequent targets to redefine this if needed -->
+	<Target Name="CompileDesignTime" />
+
+	<!-- Rust specific props -->
+	<PropertyGroup>
+		<AuthMode Condition="'$(AuthMode)' == '' and '$(Q_SESSION_GUID)' != ''">false</AuthMode>
```
dfederm, line 82: Avoid exposing CloudBuild-specific variables publicly

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+		<AuthMode Condition="'$(AuthMode)' == '' and '$(TF_BUILD)' != ''">false</AuthMode>
+		<AuthMode Condition="'$(AuthMode)' == ''">true</AuthMode>
```
dfederm, line 84: AuthMode seems unused?

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+		<CargoRegistryFeedName Condition="'$(CargoRegistryFeedName)' == ''">cargo.io</CargoRegistryFeedName>
```
dfederm, line 85: Is this the correct place the feed should be defined? Shouldn't it be defined in whatever the NuGet.config equivalent is for Rust?

src/CargoBuild/sdk/Sdk.props
```diff
@@ -0,0 +1,94 @@
...
+		<BuildCommandArgs Condition="'$(BuildCommandArgs)' == ''"></BuildCommandArgs>
+		<TestCommandArgs Condition="'$(TestCommandArgs)' == ''"></TestCommandArgs>
+		<CleanCommandArgs Condition="'$(CleanCommandArgs)' == ''"></CleanCommandArgs>
+		<RunCommandArgs Condition="'$(RunCommandArgs)' == ''"></RunCommandArgs>
+		<DocCommandArgs Condition="'$(DocCommandArgs)' == ''"></DocCommandArgs>
+		<!--For internal Microsoft use only, ignore.-->
```
dfederm, line 91: Who is the "ignore" directive for? Not sure it makes sense to have this part of the comment.

src/CargoBuild/sdk/Sdk.targets
```diff
@@ -0,0 +1,130 @@
...
+    We can't default LanguageTargets it is set in the SDK and immediately imported.  So we can only default
+    it if we know the SDK won't.  Projects probably won't load in Visual Studio but will build from the
+    command-line just fine.
+  -->
+	<PropertyGroup>
+		<LanguageTargets Condition=" '$(LanguageTargets)' == '' And '$(MSBuildProjectExtension)' != '.csproj' And '$(MSBuildProjectExtension)' != '.vbproj' And '$(MSBuildProjectExtension)' != '.fsproj' ">$(MSBuildToolsPath)\Microsoft.Common.targets</LanguageTargets>
+		<MSBuildAllProjects Condition="'$(MSBuildToolsVersion)' != 'Current'">$(MSBuildAllProjects);$(MsBuildThisFileFullPath)</MSBuildAllProjects>
+	</PropertyGroup>
+
+	<Import Project="$(CustomBeforeNoTargets)" Condition="'$(CustomBeforeNoTargets)' != '' and Exists('$(CustomBeforeNoTargets)')" />
```
dfederm, line 22: Fix property name

src/CargoBuild/sdk/Sdk.targets
```diff
@@ -0,0 +1,130 @@
...
+  -->
+	<Target Name="GetTargetPathWithTargetPlatformMoniker" />
+
+	<!--
+    The GetReferenceAssemblyPaths does not need to run since reference assemblies aren't needed.
+  -->
+	<Target Name="GetFrameworkPaths" DependsOnTargets="$(GetFrameworkPathsDependsOn)" />
+	<Target Name="GetReferenceAssemblyPaths" DependsOnTargets="$(GetReferenceAssemblyPathsDependsOn)" />
+
+	<Import Project="$(CustomAfterNoTargets)" Condition="'$(CustomAfterNoTargets)' != '' and Exists('$(CustomAfterNoTargets)')" />
```
dfederm, line 87: Fix property name

src/CargoBuild/sdk/Sdk.targets
```diff
@@ -0,0 +1,130 @@
...
+	</Target>
+	<Target Name="Clean" BeforeTargets="CargoBuild" Condition="$(Clean)=='true'">
+		<CargoBuildTask StartupProj="$(MSBuildProjectFullPath)" Command="clean" CommandArgs="$(CleanCommandArgs)" />
+	</Target>
+	<Target Name="Run" AfterTargets="CargoBuild" Condition="$(Run)=='true'">
+		<CargoBuildTask StartupProj="$(MSBuildProjectFullPath)" Command="run" CommandArgs="$(RunCommandArgs)" />
+	</Target>
+	<Target Name="Doc" AfterTargets="CargoBuild" Condition="$(Doc)=='true'">
+		<CargoBuildTask StartupProj="$(MSBuildProjectFullPath)" Command="doc" CommandArgs="$(DocCommandArgs)" />
+	</Target>
```
dfederm, line 126: For these conditions, where do these properties come from? Why are there conditions at all?

Shouldn't the `AfterTargets` be a `DependsOnTargets`?

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+using Tasks = System.Threading.Tasks;
+
+namespace MSBuild.CargoBuild
+{
+    /// <summary>
+    /// Builds rust projects using cargo.
+    /// </summary>
+    public class CargoBuildTask : Task
+    {
+        private static readonly string _tempPath = $"{Environment.GetEnvironmentVariable("TEMP")}";
```
dfederm, line 26: Interpolation is not required here

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                return (ExitCode)await processTask;
+            }
+            catch (Exception ex)
+            {
+                Log.LogWarningFromException(ex);
+                return await Tasks.Task.FromResult(ExitCode.Failed);
+            }
+        }
+
+        private async Task<bool> DownloadRustupAsync()
```
dfederm, line 326: Shouldn't this be attached to the Restore? Network Isolation / Firewall requirements forbid the build itself from hitting the network.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+            if (Command.Equals("clearcargocache", StringComparison.InvariantCultureIgnoreCase))
+            {
+                if (Directory.Exists(_cargoHome))
+                {
+                    Log.LogMessage(MessageImportance.Normal, $"Clearing cargo cache at {_cargoHome}");
+                    Directory.Delete(_cargoHome, true);
+                }
+            }
+
+            if (!Command.Equals("fetch", StringComparison.InvariantCultureIgnoreCase))
```
dfederm, line 100: For clarity, invert this `if`. ie, if the command _does_ equal fetch, then fetch. Else, do the other thing.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+            {
+                var dir = Directory.GetParent(StartupProj!);
```
dfederm, line 102: Why is the `!` needed? It's not marked as nullable.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                bool cargoFile = File.Exists(Path.Combine(dir!.FullName, "cargo.toml"));
```
dfederm, line 103: `dir!.FullName` isn't safe and may thrown an exception.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+
+                return !cargoFile || CargoRunCommandAsync(Command.ToLower(), CommandArgs).GetAwaiter().GetResult() == ExitCode.Succeeded;
```
dfederm, line 105: This should log an explicit error when the cargo file doesn't exist. What you have now will just fail the task with no error for the user.

dfederm: Actually, if the file doesn't exist, it returns `true` with no error? This doesn't seem correct.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+            return await ExecuteProcessAsync(_cargoPath, $"{command} {args}", ".", _envVars);
+        }
+
+        private async Task<bool> FetchCratesAsync(string project)
+        {
+            var stopwatch = new Stopwatch();
+            try
+            {
+                stopwatch.Start();
+                Log.LogMessage(MessageImportance.Normal, $"---- CargoBuild fetch Starting ----\n\n");
```
dfederm, line 123: This formatting is odd.

Also, the line endings should be more generic.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                foreach (var projects in rustProjects)
+                {
+                    string path = projects;
+
+                    var fetchTask = RustFetchAsync(path, EnableAuth);
+                    tasks.Add(fetchTask);
+                }
+
+                await Tasks.Task.WhenAll(tasks.ToArray());
+                bool success = tasks.Select(x => (int)x.Result).Sum() == 0;
```
dfederm, line 180: This can overflow. Also, exit codes could add up to 0 despite being errors (-1 and 1).

Instead this should check if any are non-zero.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+
+                foreach (var projects in rustProjects)
+                {
+                    string path = projects;
+
+                    var fetchTask = RustFetchAsync(path, EnableAuth);
+                    tasks.Add(fetchTask);
+                }
+
+                await Tasks.Task.WhenAll(tasks.ToArray());
```
dfederm, line 179: You shouldn't need this `ToArray`

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                if (_shouldCleanRustPath)
+                {
+                    CleanupRustPath();
+                }
+            }
+        }
+
+        private async Task<ExitCode> RustFetchAsync(string workingDir, bool authorize = false)
+        {
+            ExitCode authResult = authorize ? await DoRegistryAuthAsync(workingDir) : ExitCode.Succeeded;
```
dfederm, line 224: Seems like this would `DoRegistryAuthAsync` N times, once per project. Can they be grouped by feed?

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+
+            if (authorize && authResult == ExitCode.Succeeded || !authorize)
```
dfederm, line 226: This is more simply `authResult == ExitCode.Succeeded`

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                Log.LogMessage(MessageImportance.Normal, $"Fetching cargo crates for project in {workingDir}");
+                var exitCode = await ExecuteWithRetriesAsync(path, processArgs: args, workingDir, retryCount: RetryCnt, processRetryArgs: args);
+                Log.LogMessage(MessageImportance.Normal, $"Finished fetching cargo crates for project in  {workingDir}");
+                return exitCode;
+            }
+
+            return authResult;
+        }
+
+        private async Task<ExitCode> ExecuteWithRetriesAsync(string processFileName, string processArgs, string workingDir, int retryCount, string? processRetryArgs = null)
```
dfederm, line 244: What're the retries for?

Other SDKs don't do the same thing. Like for a csproj if you have a syntax error it doesn't just try again a bunch of times.

Are we expecting inherent flakiness with cargo?

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                await Tasks.Task.Delay(wait);
+                exitCode = await ExecuteProcessAsync(processFileName, processRetryArgs ?? processArgs, workingDir);
+            }
+
+            return exitCode;
+        }
+
+        private async Task<ExitCode> DoRegistryAuthAsync(string workingDir)
+        {
+            return await ExecuteWithRetriesAsync(_cargoPath, $"login --registry {RegistryFeedName}", workingDir, retryCount: 2);
```
dfederm, line 264: How does auth work with cargo? Will this pop up an auth window? Does it require user input on the console?

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                        {
+                            if (!info.EnvironmentVariables.ContainsKey(envar.Key))
+                            {
+                                info.EnvironmentVariables.Add(envar.Key, envar.Value);
+                            }
+                        }
+                    }
+
+                    var process = Process.Start(info);
+                    int maxWait = 1000 * 60 * 15;
```
dfederm, line 294: Do we need a timeout? Other SDKs, eg for .NET, do not enforce a timeout on the compiler

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+
+                    var nameAndExtension = fileName.Substring(fileName.LastIndexOf('\\') + 1);
+                    Log.LogMessage(MessageImportance.Normal, $"\t\t\t\t\t\t*********** Start {nameAndExtension} logs ***********\n\n");
```
dfederm, line 297: Odd formatting choice. Do we need this?

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                        process.Kill();
+                        Log.LogError($"Killed process after max timeout reached : '{info.FileName}'");
+                        return -1;
+                    }
+
+                    return process.ExitCode;
+                });
+
+                processTask.Start();
+                return (ExitCode)await processTask;
```
dfederm, line 317: The only valid values for ExitCode are 0 and 1, but the process's exit code could be anything. In particular, you explicitly return `-1` above in some cases.

You should do something more like:
```cs
int exitCode = await processTask;
return exitCode == 0 ? ExitCode.Successful : ExitCode.Fail;
```

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+            }
+            catch (Exception ex)
+            {
+                Log.LogWarningFromException(ex);
+                return await Tasks.Task.FromResult(ExitCode.Failed);
```
dfederm, line 322: You don't need to create a new task for this. Just `return ExitCode.Failed`

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+            }
+        }
+
+        private async Task<bool> DownloadRustupAsync()
+        {
+            if (File.Exists(_rustUpInitBinary) && await VerifyInitHashAsync())
+            {
+                return true;
+            }
+            else if (File.Exists(_rustUpInitBinary))
```
dfederm, line 332: Avoid checking a 2nd time. Use a local variable

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+            {
+                // If the hash doesn't match, that likely means there is a new version of rustup-init.exe available.
+                File.Delete(_rustUpInitBinary);
+            }
+
+            string rustupDownloadLink = _rustupDownloadLink;
+            Log.LogMessage(MessageImportance.Normal, $"Downloading -- {rustupDownloadLink}");
+            using (var client = new HttpClient())
+            {
+                Task<HttpResponseMessage> response = client.GetAsync(rustupDownloadLink);
```
dfederm, line 342: `await` here?

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                if (!Directory.Exists(_rustInstallPath))
+                {
+                    Directory.CreateDirectory(_rustInstallPath);
+                }
+
+                using var responseStream = new FileStream(_rustUpInitBinary, FileMode.CreateNew);
+                HttpResponseMessage res = await response;
```
dfederm, line 349: You will want to `response.EnsureSuccessStatusCode();`

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+            Log.LogMessage(MessageImportance.Normal, $"Downloading -- {rustupDownloadLink}");
+            using (var client = new HttpClient())
+            {
+                Task<HttpResponseMessage> response = client.GetAsync(rustupDownloadLink);
+                if (!Directory.Exists(_rustInstallPath))
+                {
+                    Directory.CreateDirectory(_rustInstallPath);
+                }
+
+                using var responseStream = new FileStream(_rustUpInitBinary, FileMode.CreateNew);
```
dfederm, line 348: Rename, as this isn't the http response stream.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                HttpResponseMessage res = await response;
+                await res.Content.CopyToAsync(responseStream);
+            }
+
+            return await VerifyInitHashAsync();
```
dfederm, line 353: I don't understand this part. You downloaded the file, so what hash are you verifying against?

dfederm, line 353: There's a race here since you're not downloading the files at the same time.

Should we retry this specific operation on failure?

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+        }
+
+        private async Task<bool> InstallRust()
```
dfederm, line 356: Method name should have "Async" suffix

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+        {
+            var rustupBinary = UseMsRustUp ? _msRustupBinary : _rustUpBinary;
+            if (File.Exists(_cargoPath) && Directory.Exists(_rustupHome) && File.Exists(_rustUpBinary) || File.Exists(_cargoHome) && UseMsRustUp != true || File.Exists(_msRustupBinary))
```
dfederm, line 359: Use parens as I can't tell what's required and what is "or'd"

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                {
+                    Log.LogMessage(MessageImportance.Normal, "Installing MS Rustup");
+                    string sdkRootPath = Path.Combine(workingDirPart!, "content\\dist");
+                    ExecuteProcessAsync("powershell.exe", $".\\msrustup.ps1 '{_cargoHomeBin}'", sdkRootPath, _envVars).GetAwaiter().GetResult();
+                }
+            }
+
+            foreach (var envVar in _envVars)
+            {
+                Environment.SetEnvironmentVariable(envVar.Key, envVar.Value);
```
dfederm, line 377: This changes the env vars for the MSBuild.exe process. This should be avoided.

Instead, change the environment of the child process you're launching.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+            }
+
+            ExitCode exitCode = ExitCode.Succeeded;
+            ExitCode exitCodeToolChainLatest = ExitCode.Succeeded;
+            ExitCode exitCodeLatest = ExitCode.Succeeded;
+
+            if (UseMsRustUp && File.Exists("rust-toolchain.toml"))
```
dfederm, line 384: It's unclear what this path is relative to at this point.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+            if (!string.IsNullOrEmpty(_currentRustupInitExeCheckSum))
+            {
+                return _currentRustupInitExeCheckSum!;
+            }
+
+            using var client = new HttpClient();
+            string response = await client.GetStringAsync(_checkSumVerifyUrl);
+            _currentRustupInitExeCheckSum = response.Split('\n')[0];
+
+            _currentRustupInitExeCheckSum = _currentRustupInitExeCheckSum!.ToUpperInvariant();
```
dfederm, line 425: Why `ToUpperInvariant`? Avoid the allocation

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+            using FileStream stream = File.OpenRead(_rustUpInitBinary);
+            byte[] hash = sha256.ComputeHash(stream);
+            string converted = BitConverter.ToString(hash);
+
+#if NETFRAMEWORK
+            converted = converted.Replace("-", string.Empty);
+#else
+            converted = converted.Replace("-", string.Empty, StringComparison.Ordinal);
+#endif
+            return converted == await GetHashAsync();
```
dfederm, line 411: Use `.Equals` for string comparisons.

src/CargoBuild/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+1) in your global.json
+```json
+ "msbuild-sdks": {
+    ...,
+    "Microsoft.Build.CargoBuild": "1.0.270-gf406f8eaa0"
+  },
+```
+
+2) foreach rust project a .cargosproj project file at the same level as your cargo.toml file. The project file should import the CargoBuild sdk.
```
dfederm, line 14: "cargosproj" or "cargoproj"? I'm not sure what the "s" is for.

If you were following csproj, the "cs" there is for CSharp.

dfederm, line 14: >import the CargoBuild sdk

Is "import" the right verb? It makes me think of the MSBuild `<Import>`, which is not how you use MSBuild SDKs.

src/CargoBuild/README.md
```diff
@@ -0,0 +1,121 @@
...
+```xml
+<Project Sdk="Microsoft.Build.CargoBuild">
+  <PropertyGroup>
+    <TargetFramework> your framework version </TargetFramework>
```
dfederm, line 18: Any guidance for what `TargetFramework` to use? It's not clear what the target framework for Rust would be.

src/CargoBuild/README.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+### Usage
+To restore rust dependencies, you can use the following msbuild command:
+```shell
+msbuild /t:Restore
+```
+or 
+```shell
+msbuild /restore
```
dfederm, line 30: `msbuild /restore` does a restore and a build, not just a restore.

src/CargoBuild/README.md
```diff
@@ -0,0 +1,121 @@
...
+``` 
+
+To build a rust project, you can use the following msbuild command:
+```shell
+msbuild /t:Build
```
dfederm, line 35: Or just `msbuild`?

src/CargoBuild/README.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+To clean a rust project, you can use the following msbuild command:
+```shell
+msbuild /p:clean=true
```
dfederm, line 40: Why doesn't this follow convention? Should be `msbuild /t:clean`

dfederm: Similarly with the commands below.

src/CargoBuild/README.md
```diff
@@ -0,0 +1,121 @@
...
+Specifically an itemgroup with a None item for the rust source files.
+
+```xml
+<Project Sdk="Microsoft.Build.CargoBuild">
+  <PropertyGroup>
+    <TargetFrameworks> your framework version </TargetFrameworks>
+      <OutputDir>target</OutputDir>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Include="src/*.rs" />
```
dfederm, line 74: As mentioned above, why not just let the default items pick this up? You can set the language extension to be `.rs`

src/CargoBuild/README.md
```diff
@@ -0,0 +1,121 @@
...
+  </ItemGroup>
+</Project>
+```
+Next you will need to run [slngen](https://github.com/microsoft/slngen) to generate the .sln file for the project.
```
dfederm, line 78: Is slngen a requirement to use cargoproj? I'm not sure why it's mentioned here.

src/CargoBuild.UnitTests/CargoBuildTest.cs
```diff
@@ -0,0 +1,424 @@
...
+                    projectReferenceTargets.Where(i => i.EvaluatedInclude.Equals(target, StringComparison.OrdinalIgnoreCase))
+                        .Select(i => i.GetMetadata("Targets")?.EvaluatedValue)
+                        .Where(t => !string.IsNullOrEmpty(t));
+
+                buildTargets.ShouldNotBeEmpty();
+            }
+        }
+
+        [Fact]
+        public void ProjectsCanDependOnCargoBuildProjects()
```
dfederm, line 148: Isn't this backwards? The cargo projects are depending on the csproj.

Also, I don't see anything in the SDK which indicates that a ProjectReference does anything but enforce a build ordering. I thought we had talked about, for example, a vcxproj referencing a rust project and being able to wire that up correctly? I'm not seeing anything in the SDK that would make that work properly.

CZEMacLeod:
Might have already been discussed, and I don't really use rust, but it seems odd to have build in the name twice.
`Microsoft.Build.Rust.Cargo' or similar would make it much more obvious what this sdk is about.
Again for the project type extension - given vbproj, csproj, etc. perhaps rsproj would make more sense.
Just some food for thought...

AndyGerlicher:
> Might have already been discussed, and I don't really use rust, but it seems odd to have build in the name twice. `Microsoft.Build.Rust.Cargo' or similar would make it much more obvious what this sdk is about. Again for the project type extension - given vbproj, csproj, etc. perhaps rsproj would make more sense. Just some food for thought...

Good feedback. We did discuss this a bit. Originally had Rust in the name and `rsproj`, but the decision was to focus on the integration between MSBuild and Cargo. Having it too focused on Rust (or `.rsproj`) implies more language integration and that there might be Visual Studio support, which is not what this is about. This is just a bridge between MSBuild and Cargo so you can have a single graph and get the benefits of that (caching, etc.). Make more sense given the additional context?

CZEMacLeod:
> Good feedback. We did discuss this a bit. Originally had Rust in the name and `rsproj`, but the decision was to focus on the integration between MSBuild and Cargo. Having it too focused on Rust (or `.rsproj`) implies more language integration and that there might be Visual Studio support, which is not what this is about. This is just a bridge between MSBuild and Cargo so you can have a single graph and get the benefits of that (caching, etc.). Make more sense given the additional context?

Fair enough. From an outsider's perspective it seems a little odd - given in the future that an LSP might be available, or you might open a solution including this with VSCode, where better rust integration could be available.
I understand that this isn't first party for VisualStudio though so .cargoproj makes some sense.
To me, if you are focusing on that though, the name should be aligned through-out: .*cargo*proj and Microsoft.Build.*Cargo*. Repeating the `Build` just looked weird.

Does this only output Nuget packages for libraries, or can it build executables (such that it could be referenced by Aspire)?

How does this compare to the Android C++/Java Gradle build project build which is about the closest thing I can think of.

Anyway, it looks like you've already thought about this so I'll but out, and leave you to finish this impressive bit of code.

src/CargoBuild/CargoBuildTask.cs
```diff
@@ -0,0 +1,437 @@
...
+                graphLoadStopWatch.Start();
+
+                var evaluationContext = EvaluationContext.Create(EvaluationContext.SharingPolicy.Shared);
+
+                var graph = new ProjectGraph(
+                    [new ProjectGraphEntryPoint(project)],
+                    ProjectCollection.GlobalProjectCollection,
+                    (string projectPath, Dictionary<string, string> globalProperties, ProjectCollection projCollection) =>
+                    {
+                        var loadSettings = ProjectLoadSettings.DoNotEvaluateElementsWithFalseCondition;
```
dfederm, line 134: Use `ProjectLoadSettings.IgnoreEmptyImports | ProjectLoadSettings.IgnoreInvalidImports | ProjectLoadSettings.IgnoreMissingImports | ProjectLoadSettings.DoNotEvaluateElementsWithFalseCondition`

Since this happens during the restore, Pkg variables will not be defined yet so imports of files inside packages will fail.

This uses the same flags that NuGet itself does during restore to load projects.

src/CargoBuild/CargoBuildTask.cs
novacole: In practice, auth would only happen once per feed.

src/CargoBuild/README.md
novacole: that was a typo

AndyGerlicher (dismissed):

AndyGerlicher (approved):

PR #601, opened 2025-02-04
Ensure installation occurs exactly once, add msrustup file env var 

novacole:
This change updates the sdk to ensure that cargo and rust, are installed exactly once. It also adds further support for msrustup auth with the addition of the msrustup_file env var

### DIFF
--- a/src/Cargo/sdk/CargoInstall.proj
+++ b/src/Cargo/sdk/CargoInstall.proj
@@ -1,13 +1,8 @@
 <Project>
-  <Import Project="$(MSBuildThisFileDirectory)sdk.props" />
-  <UsingTask TaskName="Microsoft.Build.Cargo.CargoTask"
-    AssemblyFile="$(MSBuildThisFileDirectory)..\build\net8.0\Microsoft.Build.Cargo.dll"
-    Condition="'$(MSBuildRuntimeType)' == 'Core' And '$(ShouldImportSdkDll)' != 'false'" />
-  <UsingTask TaskName="Microsoft.Build.Cargo.CargoTask"
-    AssemblyFile="$(MSBuildThisFileDirectory)..\build\net472\Microsoft.Build.Cargo.dll"
-    Condition="'$(MSBuildRuntimeType)' != 'Core' And '$(ShouldImportSdkDll)' != 'false'" />
+  <Import Project="$(MSBuildThisFileDirectory)Sdk.props"/>
+  <UsingTask TaskName="Microsoft.Build.Cargo.CargoTask" AssemblyFile="$(MSBuildThisFileDirectory)..\build\net8.0\Microsoft.Build.Cargo.dll" Condition="'$(MSBuildRuntimeType)' == 'Core'" />
+  <UsingTask TaskName="Microsoft.Build.Cargo.CargoTask" AssemblyFile="$(MSBuildThisFileDirectory)..\build\net472\Microsoft.Build.Cargo.dll" Condition="'$(MSBuildRuntimeType)' != 'Core'" />
   <Target Name="CargoInstall">
-    <CargoTask EnableAuth="$(AuthMode)" StartupProj="$(MSBuildStartupDirectory)" Command="install"
-      UseMsRustUp="$(UseMsRustUp)" />
+    <CargoTask EnableAuth="$(AuthMode)" StartupProj="$(MSBuildStartupDirectory)" Command="install" UseMsRustUp="$(UseMsRustUp)"/>
   </Target>
 </Project>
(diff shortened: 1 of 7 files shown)


src/Cargo/sdk/CargoInstall.proj
```diff
@@ -0,0 +1,13 @@
+<Project>
+  <Import Project="$(MSBuildThisFileDirectory)sdk.props" />
```
AndyGerlicher, line 2: nit: `Sdk.props`

src/Cargo/sdk/Sdk.props
```diff
@@ -6,7 +6,8 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <UsingMicrosoftCargoSdk>true</UsingMicrosoftCargoSdk>
-    <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)' != 'Current'">$(MSBuildAllProjects);$(MsBuildThisFileFullPath)</MSBuildAllProjects>
+    <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)' != 'Current'">
```
AndyGerlicher, line 9: `MSBuildAllProjects` Is not needed anymore (unless you wanted to add a _different_ file to it. It's automatically produced by the engine and optimized. Just delete it.

src/Cargo/sdk/Sdk.props
```diff
@@ -15,24 +16,30 @@
...
     <!--
-      NuGet should always restore Cargo projects with "PackageReference" style restore.  Setting this property will cause the right thing to happen even if there aren't any PackageReference items in the project.
+      NuGet should always restore Cargo projects with "PackageReference" style restore.  Setting this
+    property will cause the right thing to happen even if there aren't any PackageReference items in
+    the project.
     -->
     <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
 
     <!-- Targeting packs shouldn't be referenced as Cargo projects don't compile .net code. -->
-    <DisableImplicitFrameworkReferences Condition="'$(DisableImplicitFrameworkReferences)' == ''">true</DisableImplicitFrameworkReferences>
```
AndyGerlicher, line 31: I don't know about these. This is changing the value of the property. I'm pretty sure that MSBuild trims this, but the whitespace will be included in the raw value.

src/Cargo/sdk/Sdk.targets
```diff
@@ -86,47 +98,71 @@
...
-  <UsingTask TaskName="Microsoft.Build.Cargo.CargoTask" AssemblyFile="$(MSBuildThisFileDirectory)..\build\net472\Microsoft.Build.Cargo.dll" Condition="'$(MSBuildRuntimeType)' != 'Core' And '$(ShouldImportSdkDll)' != 'false'"/>
-  <Target Name="CargoFetch" AfterTargets="Restore">
-    <CargoTask EnableAuth="$(AuthMode)" StartupProj="$(MSBuildProjectFullPath)" Command="fetch" />
+  <UsingTask TaskName="Microsoft.Build.Cargo.CargoTask"
+    AssemblyFile="$(MSBuildThisFileDirectory)..\build\net8.0\Microsoft.Build.Cargo.dll"
+    Condition="'$(MSBuildRuntimeType)' == 'Core' And '$(ShouldImportSdkDll)' != 'false'" />
+  <UsingTask TaskName="Microsoft.Build.Cargo.CargoTask"
+    AssemblyFile="$(MSBuildThisFileDirectory)..\build\net472\Microsoft.Build.Cargo.dll"
+    Condition="'$(MSBuildRuntimeType)' != 'Core' And '$(ShouldImportSdkDll)' != 'false'" />
+  <Target Name="CargoInstall" AfterTargets="_CollectRestoreInputs">
```
AndyGerlicher, line 133: @jeffkl Does this look ok for `AfterTargets` to execute in Restore? Since `AfterTargets="Restore"` won't work in static graph restore. Hate to reference an _ target, but we didn't see another option.

AndyGerlicher: Also, would like your input on the `RemoveProperties` if you think we need anything else to de-dupe this.

jeffkl: Yeah so `_CollectRestoreInputs` is run for regular command-line restore and Visual Studio design-time build, while `Restore` is run for static graph-based command-line restores as well as regular restores.  But `Restore` is not run in Visual Studio, just `_CollectRestoreInputs`.  Be aware that we changed that logic recently so there's no guarantee that `_CollectRestoreInputs` will continue to work.

Does this need to work in Visual Studio?  If not, I would just run after `Restore`.

> Also, would like your input on the RemoveProperties if you think we need anything else to de-dupe this.

I would hate for a new property to get introduced that breaks this, I wish there was a `RemoveProperties="*"`.  Maybe @rainersigwald has a suggestion to future-proof this.

rainersigwald: Nothing for you today :( but did finally file it: https://github.com/dotnet/msbuild/issues/11390

jeffkl: Thanks for taking a look and opening an issue!

AndyGerlicher (approved):

PR #606, opened 2025-02-13
Update SDK to 9.0.201

MattKotsenas:
Update to the RTM version of the 9 SDK and remove the RC workaround.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "9.0.100",
+    "version": "9.0.201",
     "rollForward": "latestMinor",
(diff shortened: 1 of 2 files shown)


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "9.0.100-rc.1.24452.12",
+    "version": "9.0.100",
```
AndyGerlicher, line 3: Can you use `9.0.201` please

MattKotsenas: Fixed!

AndyGerlicher (approved):
Thanks for doing this!

PR #610, opened 2025-03-09
Set default TFM for Traversal to SDK version

### LINKED ISSUE #602: Publish on a Traversal target uses Debug configuration instead of Release by default
It seems that running `dotnet publish` on a traversal project uses Debug as the default configuration, whereas post .NET 8, the default configuration for publish should be release.

MattKotsenas:
Fixes #602.

Defaulting to `net45` feels odd, as it means folks are taking _new_ dependencies on .NET Framework without realizing it. It can also lead to oddities like #602 where old defaults unexpectedly apply to newer projects.

This change allows the user to specify a particular version as before. However, otherwise `$(NETCoreAppMaximumVersion)` becomes the default version.

I think this strikes a better balance, where users that care or need a specific version continue to do so, while those that consider Traversal projects more a facility of the SDK / build system get a more sensible default than a version released 13 years ago.

### DIFF
--- a/src/Traversal/Sdk/Traversal.targets
+++ b/src/Traversal/Sdk/Traversal.targets
@@ -26,9 +26,10 @@
     <EnableDefaultItems>false</EnableDefaultItems>
 
     <!--
-      TargetFramework is required for restore and used to default to .NET Framework v4.5.  However, Traversal projects don't specify a version so it needs to be defaulted here.
+      TargetFramework is required for restore. If the user specifies a version, use that. Otherwise, use the
+      NETCoreAppMaximumVersion so that the project acts as if it were part of the SDK.
     -->
-    <TargetFramework Condition="'$(TargetFramework)' == ''">net45</TargetFramework>
+    <TargetFramework Condition=" '$(TargetFramework)' == '' ">net$(NETCoreAppMaximumVersion)</TargetFramework>
   </PropertyGroup>
 
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
--- a/src/Traversal.UnitTests/TraversalTests.cs
+++ b/src/Traversal.UnitTests/TraversalTests.cs
@@ -204,8 +204,15 @@ namespace Microsoft.Build.Traversal.UnitTests
         [InlineData("RestoreProjectStyle", null, "PackageReference")]
         [InlineData("StopOnFirstFailure", "false", "false")]
         [InlineData("StopOnFirstFailure", null, "true")]
+#if NETFRAMEWORK || NET9_0 // These map how MSBuild.ProjectCreation chooses SDK versions given the TFM of the test
         [InlineData("TargetFramework", "net6.0", "net6.0")]
-        [InlineData("TargetFramework", null, "net45")]
+        [InlineData("TargetFramework", null, "net9.0")]
+#elif NET8_0
+        [InlineData("TargetFramework", "net6.0", "net6.0")]
+        [InlineData("TargetFramework", null, "net8.0")]
+#else
+#error "Update test data for the given target framework"
+#endif
         [InlineData("TraversalProjectNames", "custom.proj", "custom.proj")]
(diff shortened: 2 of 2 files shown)


baronfel (approved):

src/Traversal/Sdk/Traversal.targets
```diff
@@ -26,9 +26,10 @@
     <EnableDefaultItems>false</EnableDefaultItems>
 
     <!--
-      TargetFramework is required for restore and used to default to .NET Framework v4.5.  However, Traversal projects don't specify a version so it needs to be defaulted here.
+      TargetFramework is required for restore. If the user specifies a version, use that. Otherwise, use the
+      NETCoreAppMaximumVersion so that the project acts as if it were part of the SDK.
     -->
-    <TargetFramework Condition="'$(TargetFramework)' == ''">net45</TargetFramework>
+    <TargetFramework Condition=" '$(TargetFramework)' == '' ">net$(NETCoreAppMaximumVersion)</TargetFramework>
```
jeffkl, line 32: Just in case `NETCoreAppMaximumVersion` isn't defined.

```suggestion
    <TargetFramework Condition=" '$(TargetFramework)' == '' And '$(NETCoreAppMaximumVersion)' != ''">net$(NETCoreAppMaximumVersion)</TargetFramework>
    <TargetFramework Condition=" '$(TargetFramework)' == ''">net45</TargetFramework>
```

src/Traversal/Sdk/Traversal.targets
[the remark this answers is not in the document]
MattKotsenas: Updated!

jeffkl (approved):

src/Traversal/Sdk/Traversal.targets
MattKotsenas: I originally had that, then convinced myself that it could never be unset 😅.

Added, but for my education, do you know of any circumstances where it wouldn't be set?

baronfel: Shouldn't it always be set at this point because it's set in the 'base' SDK's Microsoft.NETCoreSdk.BundledVersions.props file, which is included by the 'base' SDK's props pass, which is imported by this SDK already?

MattKotsenas: And it is set in .props and we're reading in .targets, so it's _technically possible_...

src/Traversal/Sdk/Traversal.targets
```diff
@@ -26,9 +26,12 @@
     <EnableDefaultItems>false</EnableDefaultItems>
 
     <!--
-      TargetFramework is required for restore and used to default to .NET Framework v4.5.  However, Traversal projects don't specify a version so it needs to be defaulted here.
+      TargetFramework is required for restore. If the user specifies a version, use that. Otherwise, use the
+      NETCoreAppMaximumVersion so that the project acts as if it were part of the SDK. If for some reason that's also
+      not set, fallback to the SDK's old default of .NET Framework v4.5.
     -->
-    <TargetFramework Condition="'$(TargetFramework)' == ''">net45</TargetFramework>
+    <TargetFramework Condition=" '$(TargetFramework)' == '' And '$(NETCoreAppMaximumVersion)' != ''">net$(NETCoreAppMaximumVersion)</TargetFramework>
```
AndyGerlicher, line 33: What about something like `netstandard1.0` ?

MattKotsenas: That would be an improvement over `net45`, but it doesn't fix the linked bug. 

Targeting old netstandard means we're still opt-ed out of modern defaults like defaulting to `Release` for `publish` and `pack`.

MattKotsenas: Another alternative would be to find a solution for the project templates. Then we could default to the current SDK version in the template.

I also prefer that option a bit less since it means every Traversal would set the TFM which clutters things a bit, but it'd be less magical.

PR #616, opened 2025-04-15
Add Microsoft.Build.UniversalPackages

dfederm:
Introduces a new `Microsoft.Build.UniversalPackages` SDK which allows projects to download [Universal Packages](https://learn.microsoft.com/en-us/azure/devops/artifacts/quickstarts/universal-packages) during MSBuild's Restore target.

There's quite a bit going on here so here's the gist:
* There are 2 mechanisms to configure packages to download: `@(UniversalPackage)` items or `universal-packages.json` package list files
* Unless overridden, the ArtifactTool will be downloaded and executed to download the configured packages
* ArtifactTool unfortunately requires an auth token to be provided by env var. If the name of the env var is provided, eg in CI scenarios, that one is used. Otherwise:
  * The [Azure Artifacts Credential Provider](https://github.com/microsoft/artifacts-credprovider) is used to retrieve an auth token which is set in an env var.
  * Unless overridden, the Azure Artifacts Credential Provider will be discovered in the well-known install locations on disk or if needed the latest will be downloaded.

### DIFF
--- a/src/UniversalPackages/Microsoft.Build.UniversalPackages.csproj
+++ b/src/UniversalPackages/Microsoft.Build.UniversalPackages.csproj
@@ -1,0 +1,41 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <TargetFrameworks>net8.0;net472</TargetFrameworks>
+    <Description>Provides targets for downloading universal packages on restore.</Description>
+    <PackageTags>MSBuild MSBuildSdk universalpackages</PackageTags>
+    <ArtifactsPath>$(BaseArtifactsPath)$(MSBuildProjectName)\</ArtifactsPath>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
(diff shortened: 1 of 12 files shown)


dfederm:
Recursive restore isn't working...

src/UniversalPackages/Microsoft.Build.UniversalPackages.csproj
```diff
@@ -0,0 +1,41 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <TargetFrameworks>net8.0;net472</TargetFrameworks>
```
AndyGerlicher, line 5: You sure you don't want to make this netstandard? Huge compat savings in the future to for some light hackery.

dfederm: Multitargeting is pretty clean for MSBuild tasks and doesn't really cause issues. Just adds 1 line to choose the correct task assembly.

AndyGerlicher (approved):

AndyGerlicher (approved):

AndyGerlicher (approved):

PR #617, opened 2025-04-18
Adding an item list with the list of copied files from CopyArtifacts target

MatthieuMEZIL:
Adds the ability for users to specify an item name to place the list of copied files into which is off-by-default

```xml
<PropertyGroup>
  <ArtifactsCopiedFilesItemName>SomethingCustom</ArtifactsCopiedFilesItemName>
</PropertyGroup>
```

The Artifacts targets will then place the list of copied files in that item group.

### DIFF
--- a/src/Artifacts/build/Microsoft.Build.Artifacts.Common.targets
+++ b/src/Artifacts/build/Microsoft.Build.Artifacts.Common.targets
@@ -7,6 +7,10 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)' != 'Current'">$(MSBuildAllProjects);$(MsBuildThisFileFullPath)</MSBuildAllProjects>
+    <_ArtifactsListCopiedFiles Condition="'$(ArtifactsCopiedFiles)' != ''">true</_ArtifactsListCopiedFiles>
+    <_RobocopyListCopiedFiles Condition="'$(RobocopyCopiedFiles)' != ''">true</_RobocopyListCopiedFiles>
+    <ArtifactsCopiedFiles Condition="'$(ArtifactsCopiedFiles)' == ''">_</ArtifactsCopiedFiles>
+    <RobocopyCopiedFiles Condition="'$(RobocopyCopiedFiles)' == ''">_</RobocopyCopiedFiles>
   </PropertyGroup>
 
   <UsingTask TaskName="Robocopy"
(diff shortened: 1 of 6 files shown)


src/Artifacts/build/Microsoft.Build.Artifacts.Common.targets
```diff
@@ -34,7 +34,10 @@
       ShowErrorOnRetry="$([MSBuild]::ValueOrDefault($(ArtifactsShowErrorOnRetry), 'true'))"
       DisableCopyOnWrite="$([MSBuild]::ValueOrDefault($(DisableCopyOnWrite), 'false'))"
       Sources="@(Artifact)"
-      Condition="@(Artifact->Count()) > 0" />
+      ListCopiedFiles="$('$(Artifacts_ListCopiedFiles)' == 'true')"
+      Condition="@(Artifact->Count()) > 0">
+        <Output TaskParameter="CopiedFiles" ItemName="CopyArtifacts_CopiedFiles" />
```
jeffkl, line 39: This hardcodes the item type and I think it would work better if it was dynamic.  Instead of having a property that users set to enable the functionality, how about they specify the item type to put the list into?

```xml
<Output TaskParameter="CopiedFiles"
        ItemName="$(ArtifactsCopiedFilesOutputType)"
        Condition="'$(ArtifactsCopiedFilesOutputType)' != ''" />
```

You could also get rid of the property to enable the functionality with just:

```xml
<PropertyGroup>
  <_ArtifactsListCopiedFiles Condition="'$(ArtifactsCopiedFilesOutputType)' != ''">true</_ArtifactsListCopiedFiles>
</PropertyGroup>
```

Then pass `$(_ArtifactsListCopiedFiles)' to the `ListCopiedFiles` parameter.  Now users just have to set a property indicating what item type they want the copied files in:

```xml
<PropertyGroup>
  <ArtifactsCopiedFilesOutputType>MyCopiedArtifact</ArtifactsCopiedFilesOutputType> 
</PropertyGroup>

<Target Name="DoSomethingWithArtifacts" AfterTargets="CopyArtifacts">
  <Message Text="Copied %(MyCopiedArtifact.Identity)" />
</Target>
```

jeffkl, line 39: You'll also want to update the docs.

MatthieuMEZIL: Sounds that it is not allowed to have an Output ItemName empty even when Condition is false.

jeffkl: Hmm are you sure, I got the idea from common targets

https://github.com/dotnet/msbuild/blob/main/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2129-L2131

jeffkl (approved):

jeffkl:
@AndyGerlicher I think this PR is good to go, please review and merge/release at your earliest convenience.

AndyGerlicher (approved):

PR #618, opened 2025-05-05
AzureAuth, MSRustup and Cargo Installation improvements

novacole:
Installs AzureAuth (needed for msrustup) if it is not currently installed and msrustup is being installed locally.
Move msrustup installation to its own directory (instead of cargo_home), to avoid PATH issues
Add custom toolchain dir to PATH to avoid potential PATH issues with a pre-existing rust installation
Use feedurl in .cargo\config.toml instead of nuget.config, since many repos have separate feeds for cargo crate consumption
Adds configurable CargoOutputDir, which defaults to the project's bin directory.

### DIFF
--- a/src/Cargo/CargoTask.cs
+++ b/src/Cargo/CargoTask.cs
@@ -23,20 +23,20 @@ namespace Microsoft.Build.Cargo
     /// </summary>
     public class CargoTask : Task
     {
-        private static readonly string? _tempPath = Environment.GetEnvironmentVariable("TEMP");
-        private static readonly string _rustUpBinary = $"{_tempPath}\\cargohome\\bin\\rustup.exe";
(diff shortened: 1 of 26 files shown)


src/Cargo/CargoTask.cs
```diff
@@ -498,15 +512,40 @@ private async Task<bool> InstallRust()
             {
                 if (!_envVars.ContainsKey("MSRUSTUP_FEED_URL"))
                 {
-                    _envVars.Add("MSRUSTUP_FEED_URL", GetNugetFeedUrl() ?? string.Empty);
+                    Debugger.Launch();
```
AndyGerlicher, line 515: Left this in

novacole: removed

AndyGerlicher (approved):

PR #639, opened 2025-08-15
UniversalPackages: Add incrementality to avoid re-downloading packages

dfederm:
This change avoids downloading a universal package if the destination path already exists.

However, to combat scenarios where the package didn't download successfully (eg. a cancelled/killed restore), the path used to download the package is mapped to a temporary location then moved atomically when successful.

Here is an example showing the difference, although this example is using an extremely trivial package. A much larger package (which is where universal package shine to begin with) would produce a much more significant difference in absolute numbers.

Before:
Build 1:
image

Build 2:
image

After:
Build 1:
image

Build 2:
image

### DIFF
--- a/src/UniversalPackages/DownloadUniversalPackages.cs
+++ b/src/UniversalPackages/DownloadUniversalPackages.cs
@@ -141,9 +141,25 @@ public sealed class DownloadUniversalPackages : Task
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


mruxmohan4 (approved):

erikmav (commented):
LGTM

PR #673, opened 2026-06-17
Cargo build SDK: MSRustup scenario fixes

wheelv5:
Targeted fixes to MSRustup usage scenarios, avoiding large-scale refactoring.

### Support cross-compilation

The SDK pinned the MSRustup toolchain install to the target triple for the host architecture. That behavior is preserved by default, but new MSBuild property `MsRustupTargets` allows projects to specify which target triples to install (as a semicolon-separated list), to support cross-compilation (by `--target` arguments to `msrustup toolchain install`). E.g. to support both x64/amd64 and arm64:

   \<MsRustupTargets\>aarch64-pc-windows-msvc;x86_64-pc-windows-msvc\</MsRustupTargets\>

### Configurable Cargo build profile

The SDK forced Cargo build commands in non-Debug configurations for MSRustup projects to use a profile selection of `--<configurationName>` (expecting e.g. `--release`). That behavior is preserved by default, but new MSBuild property `MsRustupCargoProfile` causes the SDK to pass `--profile <MsRustupCargoProfile>` to Cargo instead. This enables using custom profiles defined in Cargo.toml without the side effects from using custom Configuration names, e.g.:

   \<MsRustupCargoProfile Condition="'$(Configuration)' == 'Release'"\>release-windows\</MsRustupCargoProfile\>

### Support more crate registry name variations

The authentication to crate registries for MSRustup projects leverages the environment variables `CARGO_REGISTRIES_{registryName}_TOKEN`. The variable naming requirement is that '-' in the registry name be converted to '_', which was missing from the conversion logic; it has now been added.


wheelv5:
@microsoft-github-policy-service agree

src/Cargo/CargoTask.cs
```diff
@@ -178,7 +191,7 @@ private async Task<bool> ExecuteAsync()
 
                             foreach (var registry in GetRegistries(Path.Combine(RepoRoot, _cargoConfigFilePath)))
                             {
-                                var registryName = registry.Key.Trim().ToUpper();
+                                var registryName = registry.Key.Trim().ToUpper().Replace('-', '_');
```
novacole, line 194: What is this for?

src/Cargo/README.md
```diff
@@ -78,8 +78,30 @@ msbuild /t:clearcargocache
...
\ No newline at end of file
+
+#### Optional MSRustup configuration properties
+
+ The SDK exposes a handful of MSBuild properties for advanced scenarios.
+
+##### `CargoProfile`
+
+By default the SDK derives the Cargo profile from the MSBuild `Configuration`: `Debug` uses Cargo's default debug profile, and any other configuration is
+passed as the `--<Configuration>` value (so `Release` becomes `--release`).
```
novacole: Targets like  CargoTest ,  CargoRun ,  CargoDoc , etc. now receive  CargoProfile  but still don't pass Configuration. If a user relies on Configuration-based profile detection without explicitly setting  CargoProfile , those targets will silently default to debug. Only  CargoBuild  passes Configuration

novacole (changes requested):

src/Cargo/CargoTask.cs
wheelv5: This is a requirement for how values are converted to environment variable names, very similar to how MSBuild properties are translated to environment variable names.

If the name of your crate source in the config file is "my-crates", and you try to specify the access token with environment variable "CARGO_REGISTRIES_MY-CRATES_TOKEN", it won't be found by Cargo. You need to name it "CARGO_REGISTRIES_MY_CRATES_TOKEN".

src/Cargo/README.md
wheelv5: I missed Configuration being passed in. But for the correct resolution, I need some help understanding the original intent, since it's a bit opaque to me.

Previously, as you noted, for a build command, a Debug Configuration would be "left alone", and anything else would add a "--${Configuration}" argument.

For all those other commands (test, run, install, etc.), even though "--release" and "--profile" are valid options, the task did not try to inject the profile.

Now, for those other commands, a caller could easily *specify* a profile in CommandArgs. But they could also do the same for the build command, so that doesn't help explain why the special case was necessary.

So depending on the reason, the right way to address this might be to either input Configuration for the other commands, or remove CargoProfile from the other commands.

novacole:
lgtm

dfederm (approved):

dfederm (approved):

PR #674, opened 2026-06-24
Make Robocopy task logging thread-safe to fix flaky parallel-copy test

novacole:
## Problem
The Windows CI job for #660 failed on a flaky unit test:
`RobocopyTests.SingleAndLongChainCopiesParallelCopyOriginalSourceFile(match: "*", longChainLength: 1000)` with:

```
System.IndexOutOfRangeException : Index was outside the bounds of the array.
   at System.Collections.Generic.List`1.Add(T item)
   at Microsoft.Build.Artifacts.Tasks.Robocopy.CopyFileImpl(...)  // a Log.LogMessage call
```

## Root cause
The `Robocopy` task copies files in parallel via an `ActionBlock` (`MaxDegreeOfParallelism > 1`) but invoked `Log.LogMessage`/`Log.LogError` from multiple worker threads concurrently. The test `BuildEngine` buffers logged events in a non-thread-safe `List<T>`, so concurrent `Add` calls raced and threw `IndexOutOfRangeException`. The 1000-deep parallel copy chain makes the race likely, which is why it tripped only on the Windows agent.

## Fix
Serialize all logging calls in the parallel copy path behind a dedicated `_logLock`.

## Validation
`dotnet test` on the Artifacts unit tests (the parallel-copy cases) passes across net472/net8.0/net9.0/net10.0.

### DIFF
--- a/src/Artifacts.UnitTests/RobocopyTests.cs
+++ b/src/Artifacts.UnitTests/RobocopyTests.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Build.Artifacts.UnitTests
                 Path.Combine("baz", "baz.txt"));
 
             DirectoryInfo destination = new DirectoryInfo(Path.Combine(TestRootPath, "destination"));
-            BuildEngine buildEngine = BuildEngine.Create();
+            ThreadSafeBuildEngine buildEngine = ThreadSafeBuildEngine.Create();
 
             Robocopy copyArtifacts = new Robocopy
             {
@@ -92,7 +92,7 @@ namespace Microsoft.Build.Artifacts.UnitTests
 
             DirectoryInfo destination = new DirectoryInfo(Path.Combine(TestRootPath, "destination"));
 
-            BuildEngine buildEngine = BuildEngine.Create();
+            ThreadSafeBuildEngine buildEngine = ThreadSafeBuildEngine.Create();
 
             Robocopy copyArtifacts = new Robocopy
             {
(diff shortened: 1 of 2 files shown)


src/Artifacts/Tasks/Robocopy.cs
```diff
@@ -37,6 +37,7 @@ public class Robocopy : Task
         private readonly ActionBlock<CopyJob> _copyFileBlock;
         private readonly HashSet<string> _sourceFilesEncountered = new (Artifacts.FileSystem.PathComparer);  // Reusable scratch space
         private readonly ConcurrentBag<ITaskItem> _copiedFiles = new ();
+        private readonly object _logLock = new ();
```
dfederm: If the problem is that the test BuildEngine uses a collection that isn't thread-safe, shouldn't the fix be in that mock instead of in product code? I imagine this new lock causing contention and the production code doesn't have the problem.

dfederm (changes requested):

src/Artifacts/Tasks/Robocopy.cs
CZEMacLeod: While wrapping the mocking `BuildEngine` might be a temporary fix, it seems like it will break the parallelism by forcing all logging though a single lock and result in these async tasks executing, at least partially, sequentially. This might hide potential issues which would occur in the production environment.

Perhaps a better approach would be to fix the underlying `BuildEngine` from https://github.com/jeffkl/MSBuildProjectCreator/blob/main/src/Microsoft.Build.Utilities.ProjectCreation/BuildEventArgsCollection.cs

The grouped events are stored in a `ConcurrentQueue<T>` already, only the specialized collections for errors, warnings, and messages are stored as `List<T>`.

I'm not sure why the individual event types are being stored as separate lists in this design - utilizing LINQ with `OfType<BuildErrorEventArgs>()` etc. on the single `ConcurrentQueue<BuildEventArgs>` instead might be a better approach.

dfederm (approved):

PR #676, opened 2026-08-10
Publish Cargo outputs through MSBuild

dfederm:
## Summary

This updates the Cargo SDK so generated primary outputs are published through the evaluated MSBuild `OutputPath` instead of staying only in Cargo’s target tree. That lets dependent projects consume Rust-produced artifacts through normal MSBuild project-reference contracts while keeping Cargo intermediates out of the public output directory.

The SDK still honors `CargoOutputDir` overrides and ordinary local Cargo builds. The new contract also supports explicit output declarations and automatic discovery for the common cases the SDK can safely publish.

## Why this is safe

- Cargo’s full target/intermediate tree still stays under the Cargo output directory; only selected primary outputs are staged for MSBuild consumption.
- Explicit output items override discovery, so unusual or renamed outputs can be declared without changing the default behavior.
- The publication point is integrated so outputs exist before project completion and are visible to downstream project-reference consumers.
- Package/native consumer behavior was verified, including a real VC++ project-reference scenario using a staged import library.

## Validation

| Test / build | Result |
|---|---|
| `src/Cargo/Microsoft.Build.Cargo.csproj` Release build | pass |
| Cargo unit test matrix | 280/280 passed |
| Packaged fake-Cargo integration | pass |
| Real VC++ project-reference consumer | pass |
| High-volume stream + process-tree timeout coverage | pass |
| Incremental + clean validation | pass |

## Out of scope

- This does not move the full Cargo target tree into `OutputPath`.
- This does not change downstream sample/workaround projects beyond the SDK contract itself.

### DIFF
--- a/src/Cargo.UnitTests/CargoTest.cs
+++ b/src/Cargo.UnitTests/CargoTest.cs
@@ -378,13 +378,16 @@ namespace Microsoft.Build.Cargo.UnitTests
         }
 
         [Theory]
-        [InlineData(null, null)]
-        [InlineData("cargo-target", null)]
-        [InlineData(null, @"artifacts\obj\Cargo\")]
-        public void PublishesOnlyDeclaredCargoBuildOutputs(string cargoOutputDirectoryName, string baseIntermediateOutputPath)
+        [InlineData(null, false)]
+        [InlineData("cargo-target", false)]
+        [InlineData(null, true)]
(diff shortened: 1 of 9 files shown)


novacole (approved):
lgtm

AndyGerlicher (approved):

PR #683, opened 2026-09-16
Avoid rewriting unchanged UniversalPackages props

vaboca:
## Summary

- Avoid rewriting generated `.upack.g.props` files when their contents are unchanged.
- Preserve regeneration when path properties change and stale-file deletion when no generated path properties remain.
- Add dedicated UniversalPackages tests for no-op generation, changed paths, item ordering, and referenced project graphs.

## Problem

`DownloadUniversalPackages` regenerated each project's `.upack.g.props` file during every invocation, even when the generated XML was unchanged. Advancing that file's timestamp can invalidate downstream incremental build checks.

## Fix

Build the generated XML in memory, compare it with the existing file using an ordinal text comparison, and write and log generation only when the content differs. Existing sorted property order, XML formatting, and stale-file cleanup remain unchanged.

## Validation

- `dotnet build .\MSBuildSdks.sln "/p:Platform=Any CPU" /p:Configuration=Debug /p:GeneratePackageOnBuild=false /p:UseSharedCompilation=false -v:minimal`
- The dedicated test project builds for 
et472, 
et8.0, 
et9.0, and 
et10.0 with no warnings or errors.
- 20 tests passed locally: 5 per target framework. The 
et9.0 tests used DOTNET_ROLL_FORWARD=Major because the local .NET 9 runtime is unavailable.


dfederm (dismissed):

dfederm:
Build is failing

src/UniversalPackages.UnitTests/Microsoft.Build.UniversalPackages.UnitTests.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net472;net8.0</TargetFrameworks>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="AssemblyShader">
```
dfederm: What's this for?

vaboca: It wasn't needed here. I copied it from neighboring test projects while setting up the new test project, but this suite does not require assembly shading, so I removed the reference in ae92e7e.

vaboca:
Fixed in a3a8bfe. The new test project now targets the same net472/net8.0/net9.0/net10.0 matrix as the other unit-test projects, and the full four-framework build/test flow passes locally.

dfederm (approved):

